PR #3, opened 2018-07-05
[**REVIEW**]  Implement fork() in NaCl's internal runtime

alyptik:
This is the second part of [PR #4](https://github.com/Lind-Project/Lind-GlibC/pull/4) and contains the NaCl runtime implementation of our fork() system call.

I apologize for the monstrous final commit; the changes seemed too confusing without following the whole thing through in one go, and I haven't figured out a sane way to split them yet (always open to suggestions of course).

This is a combination of both @detectivelyw and my work, so please review and let us know if you have any questions, comments or suggestions.

Also interested in comments from @vladimir-v-diaz @awwad @pkmoore @SantiagoTorres or anyone else whom I've missed.

----------

Our main alpha implementation of fork() is really just about
emulation of multiple processes using a single, sandboxed
application, NaCl.

NaCl, when coupled with Repy (our POSIX layer implemented in a safe subset
of Python), catches system calls which are aimed at the "trusted"
side of host/application code barrier.

In order to do this, the `struct NaClApp` and `struct NaClAppThread`
members have been modified in order to support parent/children
relationships, the emulation of IPC between threads as if they were
full processes and the copying of execution state, page tables, and
address space so it behaves exactly the way it would on
a modern SMP operating system kernel like Linux. Module a few of the
lower-level, bitty-gritty details such as:

	* Signal Propagation
	* TTY Sessions/Jobs

which are not fully implemented yet.

The child/parent relationships are maintained using a simple list of
children that belongs any one parent. There are two implementations
as of now: the `child_list` and `children` members, and once it becomes
bit clearer which of the two approaches is the correct one, the other
will be removed.

	* There are no changes from standard NaClApp execution until we
	  enter the `NaClSysFork()` handler in the soon to be "parent".
	  `nap->fork_state`, initially set to 0, is used as a "child or
	  parent?" predicate, so when and NaClAppThread enters the
	  syscall handler, we know which branch to send it down.

	* Execution proceeds down the parent branch first, allocates and
	  initializes a child NaClApp with data from its own members,
	  and finally calls `NaClCreateMainForkThread()`. From here it's
	  not much different from `NaClCreateMainThread()` until we arrive
	  at `NaClAppForkThreadSpawn()`.

	* The child/parent mutexes protect the shared data we will be
	  copying between parent and child threads, and the parent's
	  execution state of %rip, %rsp, %rbp, etc, is "frozen" at
	  the syscall entry point (the `struct NaClThreadContext` is
	  the representation of the architecture specific host thread
	  context holding the execution state of our _untrusted_ code).
	  We then adjust the child's stack to match the parent's stack size,
	  align it to the recommended size, and spawn our actual child
	  thread with `NaClAppThreadMake()`

	* A literal copy of the original child's thread context is made
	  that contains our TLS pointers and trampoline addresses.
      We can copy copy the _entire untrusted execution state_ from the
	  parent into the child safely with `NaClCopyExecutionContext()`.
	  This handles the logic to correctly copy our page mappings,
	  current stack data, ELF dynamic text , and then "patches" them with
	  `NaClPatchAddr()` so that any addresses which were contained in the
	  copied data are actually pointing where they should be; the child's
	  %r15 contains the address prefix we need to calculate physical
	  addresses from an NaCl page table mapping.

	* We then restore the child's trampoline addresses, TLS pointers, and
	  %r15 with the copy made previously, and set up the child's TLS slot
	  in the `nacl_user` global array of thread context pointers.
	  Eventually, we end up at `NaClAppForkThreadLauncher()` which sets
	  the child thread as as the foreground one `NaClTlsSetCurrentThread()`
	  and switches into its untrusted code.

	* The child is immediately punted back into `NaClSysFork()` since
	  that was where we left the %rip and %rsp, and since earlier
	  `nap->fork_state = 1` was set in the child's NaClApp constructor,
	  the child branch is taken this time, which simply signals a
	  condition variable, returns 0, and concludes the child copy of
	  fork().

	* The parent, who was sleeping on that condition variable, now
	  exits the `while` predicate containing `NaClXCondVarTimedWaitRelative()`
	  located at the very end of `NaClSysFork()`, returns the child's
	  "<pid>", and concludes the parent's copy of fork().

--------------

The unsquashed commits are also available for review in the [lind_project/lind_fork_unsquashed](https://github.com/Lind-Project/lind_project/tree/lind_fork_unsquashed) branch.


pkmoore:
Hey guys,

I spent some time looking through the PRs related to this and have a few comments.  I don't know a lot about the details of NaCl yet so I'm only able to comment on more language level concerns and some strategies for cleaning up "monstrous commits" :^)

First, there are frequently large blocks of commented out code throughout these commits.  If this is code that will be switched back on in the near future this is probably fine but if its going to stay commented out forever I'd just cut it out and retrieve it from version control if its needed down the line.  This would have cut a few hundred lines from the commits I think.

Second,  I see a lot of conditionals and loops without curly braces around their single-statement bodies.  If this is the programming standard we're going for then this is fine.  My personal preference is to always use them as the extra noise is worth avoiding a situation like Apple's goto fail bug of a few years ago.  I don't know if we have a defined standard around this or not.

Third,  a bunch of lines from these commits were small logging changes (i.e. moving from a NaCl logging function to the DDPRINTF macro I believe).  Its tempting to fix these as we see them but they should probably go in a separate set of commits from the work that these PRs are trying to achieve.  This would have removed a few hundred lines of noise.

Fourth,  I saw in several places macros being used to define things like MAX_PID an MAX_FD.  It might be fine in the case of NaCl but these two values are user configurable and a mismatch between the OS setting and the chosen macro value might cause issues.

Here are a few more specific things that stood out to me.  These are likely fine and just appeared weird because I lack the context required to know they're ok.

[Hard coded paths here (might be fine):](https://github.com/Lind-Project/native_client/pull/3/commits/d959c3022e257a2aa4735e0f4d22f59246994ff7#diff-faf993ab3c75b22e4ad72b464af974a3R629)

[unusual strlen() usage](https://github.com/Lind-Project/native_client/pull/3/commits/d959c3022e257a2aa4735e0f4d22f59246994ff7#diff-faf993ab3c75b22e4ad72b464af974a3R652)
Should:
strlen(path + tls_end_idx) + 1
be
strlen(path)+ tls_end_idx + 1
maybe not?

[unreachable block of code (if(0)](https://github.com/Lind-Project/native_client/pull/3/commits/d959c3022e257a2aa4735e0f4d22f59246994ff7#diff-faf993ab3c75b22e4ad72b464af974a3R1107)

[Segfaulting only under GDB is usually because of a race condition or because of ASLR being disabled under GDB in my experience.](https://github.com/Lind-Project/native_client/pull/3/commits/d959c3022e257a2aa4735e0f4d22f59246994ff7#diff-9bf3c29f927b897e07ad68999c9d607aR668)

There is a bunch of duplicated code here that could could be a pain should it need to be modified in the future.  It might be appropriate to pull these instances into their own functions so small changes don't have to be manually propagated to each repetition.

[1](https://github.com/Lind-Project/native_client/pull/3/commits/d959c3022e257a2aa4735e0f4d22f59246994ff7#diff-9bf3c29f927b897e07ad68999c9d607aR759)

[2](https://github.com/Lind-Project/native_client/pull/3/commits/d959c3022e257a2aa4735e0f4d22f59246994ff7#diff-9bf3c29f927b897e07ad68999c9d607aR896)

[3](https://github.com/Lind-Project/native_client/pull/3/commits/d959c3022e257a2aa4735e0f4d22f59246994ff7#diff-9bf3c29f927b897e07ad68999c9d607aR1027)

[4](https://github.com/Lind-Project/native_client/pull/3/commits/d959c3022e257a2aa4735e0f4d22f59246994ff7#diff-9bf3c29f927b897e07ad68999c9d607aR1190)


This is a really impressive amount of work to get done!  Good work!  I'll do my best to get up to speed with the project so I can provide more useful feedback in the future.

Thanks!

JustinCappos:
Thanks for the detailed feedback, Preston!

I'd like to reiterate the importance of some of the code clarity things
Preston mentions.  Curly braces around single line then or else clauses,
not leaving commented code around everywhere, and separation of commits to
make them easier to review (i.e., logging changes) are all really important.

alyptik:
Thanks for the feedback @pkmoore ! The commented out code and the `if (0)` was something I forgot to remove when rebasing onto lind, so I'll get to work on fixing that.

 Regarding the conditionals curly braces, after some thought, I agree we should have a standard way, so I can add them around all loop bodies, not a problem.

The logging changes, I also have no problem splitting into another PR, sure.

The weird strlen() should be equivalent, so I'll just swap it to your way since it's less mentally taxing to reason about.

The hard coded paths are not ideal,  so I'll figure out some way to make that more general.

The macros you commented about (MAX_FD and MAX_PID) are internal maximums, no relation to  system limits, so I'll rename them to make it less confusing what they are actually for.

Finally, definitely agree about the duplicated code. I'll figure out some way to generalize that as well.

Thanks for the comments Preston and Justin!

alyptik:
Revised and reopened as [PR #5](https://github.com/Lind-Project/native_client/pull/5).

PR #4, opened 2018-07-10
Add DPRINTF() and ABORT() logging wrappers

alyptik:
Add convenience macros for normal log messages:

```c
DPRINTF(detail_level, fmt, args...);
```
and for fatal log messages:

```c
ABORT(fmt, args...);
```

In addition, rename a few function parameters that shadowed file scope identifiers, and fix up the questionable code style in a couple places like the empty loop body of [L265](https://github.com/Lind-Project/native_client/compare/lind...lind_logging_changes?expand=1#diff-c4b9ff37f52d383c6ba2b523d742ebbcL265).

Implementation was based on changes requested by @pkmoore


alyptik:
I don't think this isn't really needed/useful anymore after giving it some thought; closing

PR #5, opened 2018-07-11
Add NaClSysFork() implementation

alyptik:
This is the second part of [PR #4](https://github.com/Lind-Project/Lind-GlibC/pull/4) and contains the NaCl runtime implementation of our fork() system call.

The original RFC implementation, [PR #3](https://github.com/Lind-Project/native_client/pull/3), has been split into smaller, easier to digest chunks, and much of the unrelated changes have been trimmed out. I'm also fairly sure I've addressed most of the issues @pkmoore and @JustinCappos pointed out in the RFC PR, but just let me know if there are any that I forgot.

This is a combination of both @detectivelyw and my work, so please review and let us know if you have any questions, comments or suggestions, not only from project members, but also from @vladimir-v-diaz @awwad @pkmoore @SantiagoTorres or anyone else whom I've missed.

----------

Our main alpha implementation of fork() is really just about
emulation of multiple processes using a single, sandboxed
application, NaCl.

In order to do this, the `struct NaClApp` and `struct NaClAppThread`
members have been modified in order to support parent/children
relationships, the emulation of IPC between threads as if they were
full processes, and the copying of execution state, page tables, and
address space so it behaves exactly the way it would on
a modern SMP operating system kernel like Linux (modulo a few of the
lower-level, bitty-gritty details such signal handling and TTY
sessions which are not fully implemented yet).

The child/parent ownership accounting is done using two simple
lists, one of children cage_ids ("pids") and one of pointers to
the child's NaClApp structure. Two accounting lists are kept, one
stored in the parent's NaClApp structure describing its children,
and one stored in the master (original) NaClApp structure describing
all currently running children. The master list is used to prevent
NaCl from exiting until all children have finished; this is so none of
NaClAppThreads are killed prematurely and emulate how fork() would
behave on the host system.

**Overview of NaClSysFork():**

- A child NaClApp is initialized by the parent using data
  from its own NaClApp structure and calls `NaClCreateMainForkThread()`.
  From here it's not much different from `NaClCreateMainThread()` until
  we arrive at `NaClAppForkThreadSpawn()`.

- The child/parent mutexes protect the shared data we will be
  copying between parent and child threads, and the parent's
  execution state of %rip, %rsp, %rbp, etc, is "frozen" at
  the syscall entry point (the `struct NaClThreadContext` is
  the representation of the architecture specific host thread
  context holding the execution state of our _untrusted_ code).
  We then adjust the child's stack to match the parent's stack
  size, and spawn our actual child thread with `NaClAppThreadMake()`

- A literal copy of the original child's thread context is made
  that contains our TLS pointers and trampoline addresses.
  We can copy copy the entire untrusted execution state from the
  parent into the child safely with `NaClCopyExecutionContext()`.
  This handles the logic to correctly copy our page mappings,
  current stack data, ELF dynamic text , and then "patches" them with
  `NaClPatchAddr()` so that any addresses which were contained in the
  copied data are actually pointing where they should be; the child's
  %r15 contains the address prefix we need to calculate physical
  addresses from an NaCl page table mapping.

- We then restore the child's trampoline addresses, TLS pointers, and
  %r15 with the copy made previously, change the syscall return value
  to 0, and set up the child's TLS slot in the `nacl_user` global
  array of thread context pointers. Eventually, we end up at
  `NaClAppForkThreadLauncher()` which sets the child thread as
  the foreground one with `NaClTlsSetCurrentThread()` and switches
  into its untrusted code. The child returns 0 from the syscall,
  and resumes the child's copy of the program.

- The parent simply switches back to untrusted code the same way
  it would for any other syscall, returns <pid>, and resumes the
  parent's copy of the program.


src/shared/platform/lind_platform.c
```diff
@@ -5,26 +5,35 @@
...
 #include <Python.h>
 #include <errno.h>
 
 #include "native_client/src/shared/platform/nacl_log.h"
 #include "native_client/src/shared/platform/lind_platform.h"
 
-PyObject* repylib = NULL;
-PyObject* code = NULL;
-PyObject* context = NULL;
+// yiwen
```
SantiagoTorres: I'm not sure this comment is informative. I'd rather nix instances of this or add a more descriptive string. Is this a Yiwen was here, or...?

src/shared/platform/lind_platform.c
```diff
@@ -603,74 +621,77 @@ int lind_sendto (int sockfd, size_t len, int flags, socklen_t addrlen, const str
...
-    CHECK_NOT_NULL(buf);
-    LIND_API_PART1
+    /*CHECK_NOT_NULL(dest_addr);;
+    CHECK_NOT_NULL(buf);;
+    LIND_API_PART1;
     callArgs = Py_BuildValue("(i[iiiis#s#])", LIND_safe_net_sendto, sockfd, len, addrlen, dest_addr, addrlen, buf, len);
-    LIND_API_PART2
-    LIND_API_PART3*/
+    LIND_API_PART2;
+    LIND_API_PART3*/;
```
SantiagoTorres: do we need this code here at all?

src/shared/platform/lind_platform.c
```diff
@@ -603,74 +621,77 @@ int lind_sendto (int sockfd, size_t len, int flags, socklen_t addrlen, const str
...
-    LIND_API_PART3*/
 
-    // unimplemented
+    /*
+     * LIND_API_PART1;
+     * callArgs = Py_BuildValue("(i[ii])", LIND_safe_net_getpeername, sockfd, addrlen);
+     * LIND_API_PART2;
+     * LIND_API_PART3;
+     */
+
```
SantiagoTorres: I know this is a style change, but the "unimplemented" string is gone

src/trusted/service_runtime/include/bits/nacl_syscalls.h
```diff
@@ -109,8 +109,20 @@
...
+#define NACL_sys_execv                  116
+#define NACL_sys_execve                 117
+#define NACL_sys_waitpid                118
+#define NACL_sys_dup3                   119
+#define NACL_sys_wait                   120
+
+/* jp */
+#define NACL_sys_pipe2                  121
+
+#define NACL_MAX_SYSCALLS               122
```
SantiagoTorres: You guys defined a total of 110 in MAX_SYSCALLS, but this number exceeds that. Is this intentional? If so, it's probably worth documenting.

src/trusted/service_runtime/lind_syscalls.c
```diff
@@ -5,13 +5,20 @@
...
+# undef _POSIX_C_SOURCE
+#endif
 
 #include <Python.h>
 
 #include <stdio.h>
 #include <assert.h>
 #include <sys/mman.h>
 #include <fcntl.h>
+// yiwen: added header files
```
SantiagoTorres: this comment belongs in a commit message.

src/trusted/service_runtime/lind_syscalls.c
```diff
@@ -61,142 +74,141 @@ typedef int(*CleanupType)(struct NaClApp*, uint32_t, LindArg*, void*);
...
-			free((void*)inArgs[2].ptr);
-		}
-		if(inArgs[3].ptr) {
-			free((void*)inArgs[3].ptr);
-		}
-    	free(xchangedata);
+        free((void *)inArgs[1].ptr);
+        free((void *)inArgs[2].ptr);
+        free((void *)inArgs[3].ptr);
+        free(xchangedata);
```
SantiagoTorres: :+1: for removing these NOOP guards here.

src/trusted/service_runtime/lind_syscalls.c
```diff
@@ -806,11 +859,23 @@ int32_t NaClSysLindSyscall(struct NaClAppThread *natp,
         DumpArg(&outArgSys[i]);
     }*/
 
+    // yiwen: handle lind_pipe here.
+    //        the pipe() call should initialize the pipe buffer
+    //        and returns two fds for the pipe
+    if (callNum == 61) {
+        int data[2] = {9001, 9002};
```
SantiagoTorres, line 856: I may be misreading this, but will those two fd's always be used and available?

src/trusted/service_runtime/lind_syscalls.c
```diff
@@ -806,11 +859,23 @@ int32_t NaClSysLindSyscall(struct NaClAppThread *natp,
...
+        if (!error)
+            NaClLog(LOG_ERROR, "NaClCopyOutToUser: failed! \n");
+        retval = 0;
+        goto cleanup;
+    }
+
     if(stubs[callNum].pre) {
         retval = stubs[callNum].pre(nap, inNum, inArgSys, &xchangeData);
-        if(retval) {
+        if(retval)
```
SantiagoTorres: I recall @pkmoore suggesting used braces even for single-statement if's. What's the decision there?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,239 @@
...
+
+
+/*
+ * always points at original program context
+ *
+ * -jp
+ */
+struct NaClThreadContext *master_ctx;
+
+struct NaClApp *NaClChildNapCtor(struct NaClAppThread *natp) {
```
SantiagoTorres: I think a comment describing what this function does could be useful here.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,239 @@
...
+            "or a corrupt nexe file may be responsible for this error.");
+    exit(EXIT_FAILURE);
+  }
+
+  if ((*mod_status = NaClAppPrepareToLaunch(nap_child)) != LOAD_OK) {
+    NaClLog(LOG_FATAL, "Failed to prepare child nap_parent for launch\n");
+  }
+  NaClLog(1, "Loading blob file %s\n", nap_child->nacl_file);
+  if (!nap_child->validator->readonly_text_implemented) {
+    NaClLog(LOG_FATAL, "fixed_feature_cpu_mode is not supported\n");
```
SantiagoTorres, line 114: I find it weird that `LOG_FATAL` continues execution. Is this intended?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,239 @@
...
+  }
+  NaClLog(1, "%s\n", "Enabling Fixed-Feature CPU Mode");
+  nap_child->fixed_feature_cpu_mode = 1;
+  if (!nap_child->validator->FixCPUFeatures(nap_child->cpu_features)) {
+    NaClLog(LOG_FATAL, "This CPU lacks features required by fixed-function CPU mode.\n");
+  }
+  if (!NaClAppLaunchServiceThreads(nap_child)) {
+    NaClLog(LOG_FATAL, "Launch service threads failed\n");
+  }
```
SantiagoTorres, line 123: The same to the two previous. Should these be warnings instead?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,239 @@
...
+
+  for (int old_fd = 0; old_fd < CAGING_FD_NUM; old_fd++) {
+    struct NaClDesc *old_nd;
+    int new_fd;
+    old_nd = NaClGetDesc(nap_parent, old_fd);
+    if (!old_nd) {
+      NaClLog(1, "NaClGetDesc() finished copying parent fd [%d] to child fd [%d]\n", old_fd - 1, new_fd - 1);
+      break;
+    }
```
SantiagoTorres: Would it be useful to log the opposite case?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -76,26 +301,105 @@ void WINAPI NaClAppThreadLauncher(void *state) {
...
+      case -1:
+        NaClLog(1, "Index not found in parent array: cage_id = %d\n", nap->cage_id);
+        break;
+      default:
+        nap_parent->children_ids[parent_idx] = 0;
+        NaClLog(1, "Parent new child count: %d\n", --nap_parent->num_children);
+        if (!DynArraySet(&nap_parent->children, nap->cage_id, NULL)) {
+          NaClLog(1, "Failed to remove child from parent list: cage_id = %d\n", nap->cage_id);
+        }
+      }
```
SantiagoTorres: These two blocks are pretty similar. I wonder if you can re-factor it into something more terse.

SantiagoTorres (commented):
This is an initial review. Hard to parse through a 5k~ changes  PR...

src/trusted/service_runtime/sel_main.c
```diff
@@ -185,18 +193,21 @@ int NaClSelLdrMain(int argc, char **argv) {
...
   int                           export_addr_to = -1;
+  struct NaClDesc               *blob_file = NULL;
+  char                          *blob_library_file = NULL;
 
   struct NaClApp                *nap = &state;
 
+  // argc2 and argv2 defines the NaCl file we want to run for nap2.
+  // they will be used when we try to create the thread.
+  /* int argc2; */
+  /* char **argv2; */
```
SantiagoTorres: hmm?

src/trusted/service_runtime/sel_main.c
```diff
@@ -688,16 +733,19 @@ int NaClSelLdrMain(int argc, char **argv) {
...
 
     /* Give debuggers a well known point at which xlate_base is known.  */
     NaClGdbHook(&state);
   }
 
+#if NACL_OSX
+# define _HOST_OSX 1
+#else
+# define _HOST_OSX 0
+#endif
```
SantiagoTorres, line 743: is there a reason you can't just use `NACL_OSX` rather than `_HOST_OSX`?

alyptik: Yes, it messes up linters in conditionals like https://github.com/Lind-Project/native_client/blob/lind_fork_changes/src/trusted/service_runtime/sel_main.c#L757

The conditional becomes `if (a &&)` which is invalid syntax. 

Granted, it's not super important, though. Is it better removed in your opinion?

src/trusted/service_runtime/sel_main.c
alyptik: Ah missed a spot with my squashing, good catch.

src/shared/platform/lind_platform.c
alyptik: I wasn't certain if it was some sort of standard part of our style so I just kinda followed the crowd. If it's not I'll do a quick clean to remove them then.

src/shared/platform/lind_platform.c
alyptik: I didn't really understand what this code was for honestly, so I don't know if those are placeholders for the future. I'll ask Yiwen about it tomorrow.

src/shared/platform/lind_platform.c
alyptik: Whoops that must have gotten lost in my squash. Once again I don't know how much of this commented lind platform stuff has any use, so Ill confirm with Yiwen tomorrow.

src/trusted/service_runtime/include/bits/nacl_syscalls.h
alyptik: Yeah, another part I dont really understand quite honestly either (a lot of the lind parts I still don't know very well). Will get back to you on this.

src/trusted/service_runtime/lind_syscalls.c
alyptik: Noted, will fix

src/trusted/service_runtime/lind_syscalls.c
alyptik: :)

src/trusted/service_runtime/lind_syscalls.c
alyptik: Ah another squash messup, will fix

src/trusted/service_runtime/nacl_app_thread.c
alyptik: Agreed, will fix

src/trusted/service_runtime/nacl_app_thread.c
alyptik: It doesn't, it aborts.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: This was one of the changes in my logging PR, add an ABORT macro wrapper for LOG_FATAL to make implicit abort() more obvious

src/trusted/service_runtime/nacl_app_thread.c
alyptik: It probably would be, will fix

alyptik: Hm, I probably could. I'll give it a shot.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: Hm, I probably could. I'll give it a shot.

alyptik:
All very helpful, thanks @SantiagoTorres !

src/trusted/service_runtime/lind_syscalls.c
alyptik: Just double checked on it, there were missing checks in the dup() for this case, but yes, now they are always available. Those syscalls now return `-NACL_ABI_EBADF` once you go over `FILE_DESC_MAX` which is to be decided, but is currently 256.

alyptik:
Issues addressed.

src/shared/platform/lind_platform.c
detectivelyw: This is not our standard style. I was leaving some comments of "// yiwen" when I was doing testing development of my "caging" branch to help me find code that I have added / changed. Please feel free to remove or fix these comments. Thanks!

src/shared/platform/lind_platform.c
detectivelyw: I think this is an unimplemented Lind system call, that was supposed to be redirected to our python library (SafePOSIX). And this particular function "int lind_sendto ()" seems to handle the "sendto(2)" system call. A documentation of the "sendto(2)" system call and other calls can be found here: https://lind.poly.edu/projects/project/wiki/LindSysCallManual
It might be useful to have "int lind_sendto ()" implemented later, so perhaps we could keep it and leave some notes.

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -166,6 +162,78 @@ NaClErrorCode NaClCheckAddressSpaceLayoutSanity(struct NaClApp *nap,
...
+}
+
+void NaClLogSysMemoryContent(uintptr_t sysaddr) {
+  unsigned char *addr = (unsigned char *)sysaddr;
+  UNREFERENCED_PARAMETER(addr);
+  NaClLog(1, "[Memory] Memory addr:                   %p\n", (void *)sysaddr);
+  NaClLog(1, "[Memory] Memory content (raw):          %#08lx\n", *(uint64_t *)addr);
+  NaClLog(1, "[Memory] Memory content (byte-swapped): %#08lx\n", OBJ_REP_64(addr));
+  NaClLog(1, "[Memory] Memory content (string):       %s\n", (char *)addr);
+}
```
SantiagoTorres: I wonder if making these log functions inline would help with performance

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -735,8 +798,8 @@ int NaClCreateMainThread(struct NaClApp     *nap,
     }
   }
   envv_len = 0;
-  argv_len = malloc(argc * sizeof argv_len[0]);
-  envv_len = malloc(envc * sizeof envv_len[0]);
+  argv_len = !argc ? NULL : malloc(argc * sizeof argv_len[1]);
+  envv_len = !envc ? NULL : malloc(envc * sizeof envv_len[0]);
```
SantiagoTorres: Hmm, ternary operators always confuse me. What's this achieving?

- if argv_len is null then malloc a pointer on it, otherwise make it null.
- if envv_len is null (which will always be), then malloc it, otherwise make it null.

These two lines definitely look wrong.

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -757,7 +820,7 @@ int NaClCreateMainThread(struct NaClApp     *nap,
    * data.  We are assuming that the caller is non-adversarial and the
    * code does not look like string data....
    */
-  for (i = 0; i < argc; ++i) {
+  for (i = 0; i < argc && argv; ++i) {
```
SantiagoTorres, line 730: do you mean to say `argv[i]` here? or are you checking that `argv` is not null to start with?

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -833,27 +896,27 @@ int NaClCreateMainThread(struct NaClApp     *nap,
...
+    *argloc = (uint32_t)NaClSysToUser(nap, (uintptr_t) p);
   }
 
   *p++ = 0;  /* Cleanup function pointer, always NULL.  */
   *p++ = envc;
   *p++ = argc;
 
-  for (i = 0; i < argc; ++i) {
-    *p++ = (uint32_t) NaClSysToUser(nap, (uintptr_t) strp);
+  for (i = 0; i < argc && argv; ++i) {
```
SantiagoTorres, line 813: same here.

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -904,6 +967,214 @@ int NaClCreateMainThread(struct NaClApp     *nap,
...
+  int                   i;
+  uint32_t              *p;
+  char                  *strp;
+  size_t                *argv_len;
+  size_t                *envv_len;
+  uintptr_t             stack_ptr;
+
+  retval = 0;  /* fail */
+  CHECK(argc >= 0);
+  CHECK(argv || argc);
```
SantiagoTorres: assuming the top `CHECK` passes, then this one will always pass no?

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -904,6 +967,214 @@ int NaClCreateMainThread(struct NaClApp     *nap,
...
+   * The following two loops cannot overflow.  The reason for this is
+   * that they are counting the number of bytes used to hold the
+   * NUL-terminated strings that comprise the argv and envv tables.
+   * If the entire address space consisted of just those strings, then
+   * the size variable would overflow; however, since there's the code
+   * space required to hold the code below (and we are not targetting
+   * Harvard architecture machines), at least one page holds code, not
+   * data.  We are assuming that the caller is non-adversarial and the
+   * code does not look like string data....
+   */
```
SantiagoTorres, line 933: is this a sensible assumption?

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -904,6 +967,214 @@ int NaClCreateMainThread(struct NaClApp     *nap,
...
+   * for argv and envv -- on linux, it is ARG_MAX, or 128KB -- and
+   * hence the overflow check is for obvious auditability rather than
+   * for correctness.
+   */
+  auxv_entries = 1;
+  if (0 != nap_child->user_entry_pt) {
+    auxv_entries++;
+  }
+  ptr_tbl_size = (((NACL_STACK_GETS_ARG ? 1 : 0) +
+                   (3 + argc + 1 + envc + 1 + auxv_entries * 2)) *
```
SantiagoTorres, line 967: these magic numbers should go away.

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -912,8 +933,9 @@ void NaClSetUpBootstrapChannel(struct NaClApp  *nap,
   hdr.ndescv = descs;
   hdr.ndesc_length = NACL_ARRAY_SIZE(descs);
 
-  rv = (*NACL_VTBL(NaClDesc, channel)->SendMsg)((struct NaClDesc *) channel,
-                                                &hdr, 0);
+  if (!channel)
```
SantiagoTorres: what can `NULL` channel between here and line 905?

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -1422,3 +1444,185 @@ static void StopForDebuggerInit (uintptr_t mem_start) {
...
+ */
+static void NaClVmCopyEntry(void *target_state, struct NaClVmmapEntry *entry) {
+  struct NaClApp *target = target_state;
+  uintptr_t offset = target->mem_start;
+  uintptr_t parent_offset = target->parent ? target->parent->mem_start : 0;
+  uintptr_t page_addr_child = offset + (entry->page_num << NACL_PAGESHIFT);
+  uintptr_t page_addr_parent = parent_offset + (entry->page_num << NACL_PAGESHIFT);
+  size_t copy_size = entry->npages << NACL_PAGESHIFT;
+
+  /* don't capy pages if nap has no parent */
```
SantiagoTorres: `s/capy/copy/`?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -66,17 +86,34 @@
 #include "native_client/src/trusted/service_runtime/win/debug_exception_handler.h"
 
 #if NACL_WINDOWS
-#include "native_client/src/trusted/service_runtime/win/debug_exception_handler.h"
-#include "native_client/src/shared/platform/win/xlate_system_error.h"
+#  include "native_client/src/trusted/service_runtime/win/debug_exception_handler.h"
+#  include "native_client/src/shared/platform/win/xlate_system_error.h"
```
SantiagoTorres, line 89: why?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -66,17 +86,34 @@
...
+/*
+ * yiwen: my data for the in-process-pipe
+ *
+ * char pipe_buffer[16*4096];
+ * int pipe_mutex;
+ *
+ * 0: pipe is empty, ready to write, cannot read; 1: pipe is full, ready to read, cannot write.
+ * at initialization, it should be set to 0.
+ */
+/* char *buffer_ptr; */
```
SantiagoTorres: this code is commented out. Do you need it?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -578,67 +641,100 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
...
+
+  // yiwen
+  fd = fd_cage_table[nap->cage_id][d];
+
+  /* debug -jp */
+  if (fd == 8000 || fd == 8001) {
+     NaClLog(1, "cage_id: %d fd = %d\n", nap->cage_id, fd);
+     retval = 0;
+     goto out;
+  }
```
SantiagoTorres, line 796: I would mark this block with a FIXME or a TODO in the comment.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1305,38 +1431,36 @@ int32_t NaClSysMmapIntern(struct NaClApp        *nap,
...
-  if (NULL == ndp) {
+  if (!ndp) {
     /*
      * Note: sentinel values are bigger than the NaCl module addr space.
      */
     file_size                = kMaxUsableFileSize;
     file_bytes               = kMaxUsableFileSize;
     host_rounded_file_bytes  = kMaxUsableFileSize;
-    alloc_rounded_file_bytes = kMaxUsableFileSize;
+    /* alloc_rounded_file_bytes = kMaxUsableFileSize; */
```
SantiagoTorres, line 1522: remove this line

SantiagoTorres (changes requested):
I think this is it for my first round :+1:

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: Good idea, I'll inline all the small logging functions.

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: @SantiagoTorres  It's kinda confusing but there is a point: basically it is to avoid `free(<uninitialized pointer>)` in the cleanup path. Hopefully this fix makes it clearer: 4ccc5bc

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: The second one, yes.

src/trusted/service_runtime/sel_ldr.c
alyptik: Ah, linter was giving me a false positive here about a `NULL` dereference. It also doesn't understand `LOG_FATAL` aborts :/. Will fix, good catch!

src/trusted/service_runtime/sel_ldr.c
alyptik: Indeed, haha.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Nested inside `#if NACL_WINDOWS`

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Nope, removing.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Will do

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Whoops, will fix

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: Ah yep, you are right

src/shared/platform/lind_platform.c
alyptik: Sounds good!

src/shared/platform/lind_platform.c
alyptik: Ahh perfect, noted!

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: Ah wait, `argc == 0` will pass `CHECK(argc >= 0);` but fail `CHECK(argv || argc);` if `argv == NULL`

src/trusted/service_runtime/nacl_syscall_common.c
SantiagoTorres: ah, gotcha. Thanks!

src/trusted/service_runtime/sel_main.c
SantiagoTorres: Oh no, I was just curious. This looks like a valid reason imho

src/trusted/service_runtime/lind_syscalls.c
SantiagoTorres: fair enough. I just realized that callNum is a magic number. Would be probably worth using one of the `#defines`?

src/trusted/service_runtime/nacl_app_thread.c
SantiagoTorres: fair enough :ok_hand:

src/shared/platform/lind_platform.c
SantiagoTorres: thanks for clarifying @detectivelyw !

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: good catch, will fix

src/trusted/service_runtime/sel_ldr_standard.c
SantiagoTorres: yeah, I think 4ccc5bc pretty much fixes it. Thanks!

src/trusted/service_runtime/lind_syscalls.c
alyptik: agreed, tbh the pipe fds as well, ill make enums / macros for those

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: @SantiagoTorres Turns out this is just the representation of pushing the argv, env, and auxv to the stack. The `STACK_GETS_ARG` is because on x86 it pushes an extra address due to SysV calling conventions. As esoteric as this is, there isn't really any better way to name these that I can think of.

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: It's not, but I've looked it over and fixing this part would be incredibly complex and I think I'll handle it in a separate PR.

src/trusted/service_runtime/lind_syscalls.c
alyptik: I've redone the pipe() implementation in a different branch; it's not too complicated so should I just pull it into this one?

src/trusted/service_runtime/nacl_app_thread.c
alyptik: After giving it more thought, I actually think LOG_FATAL speaks for itself. I mean, I've never seen anything that included the word fatal and didn't kill the program.

alyptik:
Issues addressed @SantiagoTorres, let me know if there's any of those fixes you disagree with. If not, I'll merge this in.

alyptik:
Closing and making cumulative PR of changes/suggestions for final review, thanks everyone.

alyptik:
Superceded by [PR #6](#6).

PR #6, opened 2018-08-28
Add userspace emulation of fork to the NaCl Internal Runtime (IRT)

alyptik:
Add a single-process implementation (using the Linux NPTL implementation
of POSIX Threads to simulate common multi-process programming) of the fork
system call. Programming and system designs leveraging this style are
ubiquitous in modern Computer Science and Networking; common implementation
of a standard web server is the single HTTP parent daemon which forks off
several children worker processes.

The main motivation for an emulation of fork (as opposed to just using
the kernel system call) is allowing for _much_ stricter control and
auditing of application(s) being run within NaCl while still being
transparent to sandboxed processes.

This provides users with a balance between useful security and practical
usability which, if not respected, tends to become "a system safe from
everyone; owner included."


alyptik:
@detectivelyw Could you please review commit [9966667b6f](native_client:9966667b6f)? These are largely your Repy related changes and would appreciate a second pair of eyes to make sure I didn't remove anything I shouldn't have.

alyptik:
@ex0dus-0x Would you mind taking a look at the `src/trusted/service_runtime/sel_*` files? Most of those changes involve the TCB between the host system and NaCl where I know your opinions would be very helpful.

alyptik:
@pkmoore Would you mind taking a look at the `src/trusted/service_runtime/nacl_*` (including the `nacl_syscall_handlers_gen.py`) files please? These contain the the actual IRT system call implementation of fork, and since you have a lot of experience at the nitty-gritty system call layer your input would be a huge help.

alyptik:
@JustinCappos @moyix I wasn't sure how busy you two were, so I didn't want to assign anything explicity for you guys. That being said, I would still very much appreciate both of your opinions if you have time.

alyptik:
I very much appreciate the help, thank you guys.

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -666,17 +680,17 @@ int NaClAddThread(struct NaClApp        *nap,
...
-    NaClLog(LOG_FATAL,
+     NaClLog(LOG_FATAL,
             "NaClRemoveThreadMu:: num_threads cannot be 0!!!\n");
   }
   --nap->num_threads;
   if (!DynArraySet(&nap->threads, thread_num, (struct NaClAppThread *) NULL)) {
-    NaClLog(LOG_FATAL,
+     NaClLog(LOG_FATAL,
             "NaClRemoveThreadMu:: DynArraySet at position %d failed\n",
             thread_num);
```
ex0dus-0x, line 694: @alyptik noticed some whitespacing inconsistencies here.

alyptik: This change was intended; the original was indented with one space incorrectly.

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -904,6 +874,215 @@ int NaClCreateMainThread(struct NaClApp     *nap,
...
+  }
+  *p++ = AT_NULL;
+  *p++ = 0;
+
+  CHECK((char *)p == (char *)stack_ptr + ptr_tbl_size);
+
+  /* now actually spawn the thread */
+  NaClXMutexLock(&nap_child->mu);
+  nap_child->running = 1;
+  NaClXMutexUnlock(&nap_child->mu);
```
ex0dus-0x, line 1050: Can you explain why the critical section here is only for setting the running flag for `nap_child`?

alyptik: We need a lock for two reasons:

1) To act as a memory barrier and synchronization point for when the parent NaClApp and functions like NaClReportExitStatus() try to query thread running state.

2) To block strange spurious thread spawns when running on a host with glibc 2.28 or later.

I am still trying to figure out the reason behind the spurious thread spawns, but I've had no luck yet. C11 threads *were* added in glibc 2.28 which may be related, but haven't found anything concrete so far.

alyptik: For what it's worth, I did copy this lock straight from the non-fork thread constructor function, so it is possible this may not be absolutely needed here.

I mostly looked at other code which read this variable and the lock seemed sane, but I'll do some testing to verify it. Good catch!

alyptik:
@aaaaalbert Whenever you have some free time, I would appreciate it if you could take a look at the `src/trusted/service_runtime/nacl_*` (including the `nacl_syscall_handlers_gen.py`) files. These contain the the actual IRT system call implementation of fork() and are the largest part, so another pair of eyes in addition to @pkmoore is probably a good idea.

@awwad I know you don't do much C, but would also appreciate just a quick glance over those files to maybe see if there are any hugely glaring mistakes I've missed.

Thanks!

src/trusted/service_runtime/posix/sel_memory.c
```diff
@@ -56,6 +56,16 @@ static int NaClPageAllocInternalFlags(void **p, size_t size, int map_flags) {
   return (NULL == addr) ? -ENOMEM : 0;
 }
 
+void *NaClPageAllocFlags(void **p, size_t size, int map_flags) {
+  if (*p) {
+      map_flags |= MAP_FIXED;
+  }
+  if (NaClPageAllocInternalFlags(p, size, map_flags) < 0) {
```
ex0dus-0x, line 63: Can you explain the functionality of this function in context of your `fork()` implementation?

alyptik: This is basically a wrapper for the internal function to allocate pages. The function itself, NaClPageAllocInternalFlags(), is the equivalent of mmap().

When we create a copy of the NaClApp during fork, we also need to make a copy of its address space. This function is what is used to allocate new pages when we walk the parent's page table for making copies of each page which doesn't refer to a file backed by persistent storage.

alyptik: I should add that the reason we need the wrapper in the first place is because `NaClPageAllocInternalFlags()` is static.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,235 @@
...
+    NaClLog(1, "Error while loading \"%s\": %s\n", nap_child->nacl_file, NaClErrorString(*mod_status));
+    NaClLog(LOG_FATAL, "%s\n%s\n",
+                       "Using the wrong type of nexe (nacl-x86-32 on an x86-64 or vice versa) ",
+                       "or a corrupt nexe file may be responsible for this error.");
+  }
+
+  if ((*mod_status = NaClAppPrepareToLaunch(nap_child)) != LOAD_OK) {
+    NaClLog(LOG_FATAL, "Failed to prepare child nap_parent for launch\n");
+  }
+  NaClLog(1, "Loading blob file %s\n", nap_child->nacl_file);
```
aaaaalbert, line 112: This (and other similar occurences of `NaClLog`) should probably use a log level macro, not a hardcoded number as its first argument.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,235 @@
...
+    natp->nap->debug_stub_callbacks->thread_create_hook(natp);
+  }
+
+#if !NACL_WINDOWS
+  /*
+   * Ensure stack alignment.  Stack pointer must be -8 mod 16 when no
+   * __m256 objects are passed (8 mod 32 if __m256), after the call.
+   * Note the current doc (as of 2009-12-09) at
+   *
+   *   http://www.x86-64.org/documentation/abi.pdf
```
aaaaalbert: Down for me.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -231,6 +511,137 @@ struct NaClAppThread *NaClAppThreadMake(struct NaClApp *nap,
...
+   */
+  nap_child->mem_start = child_ctx.r15;
+  natp_child->user.r15 = nap_child->mem_start;
+  natp_child->user.rsp = (uintptr_t)stack_ptr_child + stack_ptr_offset;
+  natp_child->user.rbp = (uintptr_t)stack_ptr_child + base_ptr_offset;
+  natp_child->user.sysret = 0;
+
+/* examine arbitrary stack values */
+#if defined(_DEBUG)
+# define NUM_STACK_VALS 16
```
aaaaalbert, line 567: That's an interesting indent for the `#if`'s "body".

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -239,17 +650,28 @@ int NaClAppThreadSpawn(struct NaClApp *nap,
...
+
   /*
    * We set host_thread_is_defined assuming, for now, that
    * NaClThreadCtor() will succeed.
    */
   natp->host_thread_is_defined = 1;
   if (!NaClThreadCtor(&natp->host_thread, NaClAppThreadLauncher, (void *) natp,
                       NACL_KERN_STACK_SIZE)) {
-    /*
+     /*
```
aaaaalbert: The edit breaks alignment of `*` with the lines below.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -302,9 +296,7 @@ int32_t NaClOpenAclCheck(struct NaClApp *nap,
     NaClLog(0, "O_RDONLY = %d\n", NACL_ABI_O_RDONLY);
     NaClLog(0, "O_WRONLY = %d\n", NACL_ABI_O_WRONLY);
     NaClLog(0, "O_RDWR   = %d\n", NACL_ABI_O_RDWR);
-#define FLOG(VAR, BIT) do {\
-      NaClLog(1, "%s: %s\n", #BIT, (VAR & BIT) ? "yes" : "no");\
-    } while (0)
+#define FLOG(VAR, BIT) NaClLog(1, "%s: %s\n", #BIT, ((VAR) & (BIT)) ? "yes" : "no")
```
aaaaalbert, line 299: This `#define` feels out of place here in the middle of the code. Move up / out to group with other log functions?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -458,54 +442,88 @@ int32_t NaClSysNameService(struct NaClAppThread *natp,
...
-  NaClLog(3, "NaClSysDup(0x%08"NACL_PRIxPTR", %d)\n",
-          (uintptr_t) natp, oldfd);
-  old_nd = NaClGetDesc(nap, oldfd);
-  if (NULL == old_nd) {
-    retval = -NACL_ABI_EBADF;
+  NaClLog(1, "NaClSysDup(0x%08"NACL_PRIxPTR", %d)\n", (uintptr_t)natp, oldfd);
+
+  if (nap->fd > FILE_DESC_MAX) {
+    ret = -NACL_ABI_EBADF;
+    goto done;
```
aaaaalbert, line 454: What's the benefit of the `goto` here and below, compared to a plain `return -NACL_ABI_...`?

aaaaalbert:
One general remark: variable names like `nap` and `natp` have a slight Hungarian touch, so to speak.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -538,30 +556,54 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
   char                 path[NACL_CONFIG_PATH_MAX];
   nacl_host_stat_t     stbuf;
   int                  allowed_flags;
+  const char           glibc_prefix[] = "/lib/glibc/";
```
aaaaalbert: Is `char []` for a string preferred over `char *`?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -538,30 +556,54 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
...
+  /*
+   * TODO:
+   * find a cleaner method to prevent the
+   * runtime linker from searching for libc
+   * in /lib/glibc/tls/ (this removes the
+   * tls/ path component).
+   *
+   * -jp
+   */
+  if (!memcmp(path, tls_prefix, sizeof tls_prefix - 1)) {
```
aaaaalbert, line 582: Possible multi-byte char encoding issue ahead?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -578,68 +620,93 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
...
+  /*
+   * register the fd and lib_path info for the cage, in lib_table[CACHED_LIB_NUM_MAX]
+   * this will be used when trying to check if a lib has been cached in our system
+   */
+  if (fd_retval > CACHED_LIB_NUM_MAX) {
+     strncpy(nap->lib_table[fd_retval].path, path, strlen(path) + 1);
+     nap->num_lib++;
+  }
+
+  NaClLog(1, "[*** Debug ***][Open] fd = %d, filepath = %s \n", fd_retval, path);
```
aaaaalbert: Besides the hardcoded log level mentioned previously, the debug message is quite different from the others I saw so far: custom prefix, contains only half of the function name; trailing space in message.

aaaaalbert:
Most functions seem to log "entering ..." and inner function calls with their parameters and return values. I wonder if there's a prettier and less verbose way to achieve this.

src/trusted/service_runtime/nacl_syscall_common.h
```diff
@@ -105,21 +110,21 @@ int32_t NaClSysStat(struct NaClAppThread *natp,
                     struct nacl_abi_stat *nasp);
 
 int32_t NaClSysMkdir(struct NaClAppThread *natp,
-                     uint32_t             path,
+                     uint32_t             pathname,
```
aaaaalbert, line 113: Does that variable name make sense? This isn't a pathname in the common meaning of the word.

src/trusted/service_runtime/nacl_syscall_common.h
```diff
@@ -301,6 +306,14 @@ int32_t NaClSysTestInfoLeak(struct NaClAppThread *natp);
 
 int32_t NaClSysTestCrash(struct NaClAppThread *natp, int crash_type);
 
+int32_t NaClSysPipe(struct NaClAppThread  *natp, uint32_t *pipedes);
+int32_t NaClSysFork(struct NaClAppThread  *natp);
+int32_t NaClSysExecv(struct NaClAppThread  *natp);
+int32_t NaClSysExecve(struct NaClAppThread  *natp, void* path, void* argv, void* envp);
```
aaaaalbert: Style consistency for pointer names (`... *natp` vs `void* ...`).

aaaaalbert:
`main` in `src/trusted/service_runtime/nacl_syscall_handlers_gen.py` isn't in great shape, and IMO we should deprecate Python 2 wherever we can.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,235 @@
...
+ * of the parents NaClApp structure which is
+ * used in NaClSysFork()
+ */
+struct NaClApp *NaClChildNapCtor(struct NaClAppThread *natp) {
+  struct NaClApp *nap_child = NaClAlignedMalloc(sizeof *nap_child, __alignof(struct NaClApp));
+  struct NaClApp *nap_master = ((struct NaClAppThread *)master_ctx)->nap;
+  struct NaClApp *nap_parent = natp->nap;
+  NaClErrorCode *mod_status = NULL;
+
+  CHECK(nap_master);
```
aaaaalbert, line 58: What `CHECK` is performed here? (I can look it up, but perhaps the macro name can guide my understanding.)

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,235 @@
...
+    NaClLog(1, "[nap %d] incrementing num_children\n", nap_cur->cage_id);
+    nap_cur->children_ids[nap_cur->num_children++] = nap_child->cage_id;
+    if (!DynArraySet(&nap_cur->children, nap_child->cage_id, nap_child)) {
+      NaClLog(LOG_FATAL, "[nap %u] failed to add cage_id %d\n", nap_cur->cage_id, nap_child->cage_id);
+    }
+    NaClLog(1, "[nap %d] new child count: %d\n", nap_cur->cage_id, nap_cur->num_children);
+    NaClXMutexUnlock(&nap_cur->children_mu);
+
+    /* break after parent's cleanup */
+    if (nap_cur == nap_parent) {
```
aaaaalbert: I don't get the combination of this `if ... break` and `nap_cur = nap_parent` in the head of the `for` loop. Can the conditional `break` just be rolled into the termination condition of the `for` loop?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,235 @@
...
+    struct NaClDesc *old_nd;
+    old_nd = NaClGetDesc(nap_parent, old_fd);
+    if (!old_nd) {
+      break;
+    }
+    new_fd = NaClSetAvail(nap_child, old_nd);
+    NaClSetDesc(nap_child, new_fd, old_nd);
+    fd_cage_table[nap_child->cage_id][new_fd] = fd_cage_table[nap_parent->cage_id][old_fd];
+    nap_child->fd++;
+    NaClLog(1, "NaClGetDesc() copied parent fd [%d] to child fd [%d]\n", old_fd, new_fd);
```
aaaaalbert, line 137: `... [%d] ...` I think introduces yet a new debug message sub-format.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -578,68 +620,93 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
...
     NaClLog(1, "NaClHostDirOpen(0x%08"NACL_PRIxPTR", %s) returned %d\n",
             (uintptr_t) hd, path, retval);
-    if (0 == retval) {
-      retval = NaClSetAvail(nap,
-                            ((struct NaClDesc *) NaClDescDirDescMake(hd)));
-      NaClLog(1, "Entered directory into open file table at %d\n",
-              retval);
+    if (!retval) {
+      retval = NaClSetAvail(nap, ((struct NaClDesc *) NaClDescDirDescMake(hd)));
+      NaClLog(1, "Entered directory into open file table at %d\n", retval);
```
aaaaalbert: Watch out for other uses of `enter` (i.e. where entering a function) in debug messages. Perhaps reword here.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -16,14 +16,235 @@
...
+}
+
+void WINAPI NaClAppForkThreadLauncher(void *state) {
+  struct NaClAppThread *natp = (struct NaClAppThread *) state;
+  struct NaClApp *nap = natp->nap;
+  struct NaClThreadContext *context = &natp->user;
+  uint32_t thread_idx;
+  nacl_reg_t secure_stack_ptr;
+
+  NaClLog(1, "%s\n", "NaClAppForkThreadLauncher: entered");
```
aaaaalbert, line 150: Inconsistent message format, compare e.g. [here](https://github.com/Lind-Project/native_client/pull/6/files#diff-4c3032ce019016f86159cf8d6d452182R62).

alyptik:
@aaaaalbert As much as I agree with you regarding the names of nap and natp, I've just been following Google's original naming convention for those to keep things consistent as that is how they are named across the entire codebase.

The function log thing is an interesting point, you are probably right. I'll see if I can figure out a way to do so.

I also agree with you on deprecating Python 2, that's something I plan to address in a future PR.

src/trusted/service_runtime/nacl_app_thread.c
aaaaalbert: This seems to do pretty much the same, including spelling out the checked condition:

```c
#include <assert.h>
int main(int argc, char** argv) {
  assert ( argc == 0 );
  return 0;
}
```
Compile and run (OS X 10.11.6, clang-800.0.42.1):
```
Assertion failed: (argc == 0), function main, file foo.c, line 3.
Abort trap: 6
```

alyptik: This is just an assert() basically, but the rest of the codebase uses CHECK() so I was keeping things consistent.

aaaaalbert:
Okay, I thought so re: `nap` and `natp`. Good to have your confirmation!

src/trusted/service_runtime/nacl_app_thread.c
alyptik: This is just to perform the loop body operations on both parent and master. I think maybe it it possible squeeze the break condition into the for statement though, let me see if I can figure out a way.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: The problem is there are 7 levels, which only are visible with that many -v flags (e.g. level 3 needs -vvv). I think adding macros for those would be a little bit too much abstraction without providing much benefit.

aaaaalbert: I just found it odd that there's a `LOG_FATAL` but then `1` to `3`.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: Yeah the debug format is a little consistent, I'll see if I can figure out a slightly cleaner way to do this.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: Same as last comment.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: Ah yep, I'll remove this from the comment.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: This indent was to keep everything consistent with the other parts of the codebase.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: Hm, maybe there is a better way to wrap these logging functions, I'll give it some thought.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: Good catch, will fix that.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Yeah, you are right, I'll hoist that out of the function.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: The gotos are to keep the error path code style consistent with the rest of the syscalls that use gotos because of their more complicated error handling.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: The glibc build always uses `/lib/glibc/tls` so it shouldn't be a problem.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Ah yeah, that is a good idea, will fix.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Yeah, I'll reword that.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Whoops, missed that. Will fix.

src/trusted/service_runtime/nacl_syscall_common.h
alyptik: This was to keep consistency with  the other system call parameter naming conventions.

src/trusted/service_runtime/nacl_syscall_common.h
alyptik: Ah, good catch. Missed this in my original revision, will fix.

aaaaalbert:
Yet another general remark: Perhaps reconsider the various platform-dependent `#ifdef`s. See section 8 of Ousterhout's [Tcl/Tk Engineering Manual](http://wfr.tcl-lang.org/fichiers/Ousterhout_Engineering_Manual.pdf) for suggestions on selecting functionality differently.

alyptik:
@aaaaalbert That's an interesting suggestion; I'll give that a read right now, thanks!

alyptik:
@aaaaalbert Pushed some commits to address your suggestions; redoing the debug messages for consistency will require a bit more thought and may have to be done in a future PR, but everything else was pretty smooth. What are your thoughts?

aaaaalbert:
94e1a351f6be08cd140ae8b4fe8cb010cff9c860 is trivially OK, though I wonder if *some* ABI docs should be referenced instead of these.

21089dae11e0f629d7b342958ec4edc64ede26c3 also looks good.

1eb5e496dc064221f3a56a32ac9544c9d7b6dc98 looks much cleaner for the removed `break`, thanks! I didn't check in great detail the other changes that this commit introduced; not sure if they are only cleanup / fallout from rewriting the loop conditions.

alyptik:
@aaaaalbert Ah, I'll see if I can find some ABI docs and will rebase if so. And regarding the break, it was indeed just that and a rename from nap_cur to cur to avoid huge lines.

alyptik:
@aaaaalbert Found a copy :)

aaaaalbert:
:+1:

src/shared/platform/lind_platform.c
```diff
@@ -5,26 +5,33 @@
...
+double lind_syscall_execution_time[NACL_MAX_SYSCALLS];
 
-static int initialized = 0;
+PyObject *py_repylib;
+PyObject *py_code;
+PyObject *py_context;
+
+static int initialized;
 
 #define REPY_RELPATH "../repy/"
```
detectivelyw: Does this path definition require corresponding file directory to be setup during installation? If so, perhaps we should provide comments and documentation to make sure things are setup correctly.

src/shared/platform/lind_platform.c
```diff
@@ -240,89 +249,94 @@ int ParseResponse(PyObject* response, int* isError, int* code, char** dataOrMess
...
+                    memcpy((var),                                       \
+                           _data + sizeof(int) * (total) + _offset,     \
+                           ((int *)_data)[(current)]);                  \
+                }                                                       \
+            }                                                           \
+            _offset += ((int*)_data)[(current)];                        \
+        }
 
-#define DUMP_DATA(x) printf(#x" = 0x%"NACL_PRIX64"\n", (uint64_t)(x));
+#define DUMP_DATA(x) printf(#x" = %#lx\n", (uint64_t)(x))
```
detectivelyw: Is there a reason for changing the NaCl default data format?

src/shared/platform/lind_platform.c
```diff
@@ -240,89 +249,94 @@ int ParseResponse(PyObject* response, int* isError, int* code, char** dataOrMess
...
+            DUMP_DATA(S_ISFIFO((x)->st_mode));                         \
+            DUMP_DATA(S_ISLNK((x)->st_mode));                          \
+            DUMP_DATA(S_ISSOCK((x)->st_mode));                         \
+            DUMP_DATA((x)->st_uid);                                    \
+            DUMP_DATA((x)->st_gid);                                    \
+            DUMP_DATA((x)->st_rdev);                                   \
+            DUMP_DATA((x)->st_size);                                   \
+            DUMP_DATA((x)->st_blksize);                                \
+            DUMP_DATA((x)->st_blocks);                                 \
+        } while (0)
```
detectivelyw: What's happening here?

src/trusted/service_runtime/lind_syscalls.c
```diff
@@ -259,46 +268,50 @@ int LindSelectPreprocess(struct NaClApp *nap, uint32_t inNum, LindArg* inArgs, v
...
                 FD_SET(hfd, (fd_set*)inArgs[3].ptr);
             }
         }
     }
-    *(int64_t*)&inArgs[0].ptr = max_hfd+1;
-    ((int*)(*xchangedata))[0] = max_hfd+1;
-    NaClLog(3, "max_fd is set to %"NACL_PRId64" was %"NACL_PRId64"\n", *(int64_t*)&inArgs[0].ptr, max_fd);
+    *(int64_t *)&inArgs[0].ptr = max_hfd + 1;
+    ((int *)(*xchangedata))[0] = max_hfd + 1;
+    NaClLog(1, "max_fd is set to %"NACL_PRId64" was %"NACL_PRId64"\n", *(int64_t *)&inArgs[0].ptr, max_fd);
```
detectivelyw, line 288: So here we are using the NaCl defined data format NACL_PRId64. I think it would be nice to keep consistence about our NaClLog data format. (See an inconsistent case before.)

src/trusted/service_runtime/lind_syscalls.c
```diff
@@ -720,95 +768,105 @@ static int NaClCopyZStr(struct NaClApp *nap,
...
-    /*
-    // For debugging
-    for(int i=0; i<(int)inNum; ++i) {
-        DumpArg(&inArgSys[i]);
+    /* yiwen: handle lind_pipe here.
+     *        the pipe() call should initialize the pipe buffer
+     *        and returns two fds for the pipe
+     */
+    if (callNum == 61) {
+        int data[2] = {9001, 9002};
```
detectivelyw: So this is my old testing code, which used hardcoded 9001 and 9002 as fds for our own pipe. If we are using different fds for our pipes now, this should be modified. If we are not using this code now, I think we should put this piece of code in a comment and mark it as "unused".

JustinCappos:
Or just delete it.  No need to junk up the code base with obsolete,
commented code when we can pull it out of github later if we really need it.

src/shared/platform/lind_platform.c
```diff
@@ -240,89 +249,94 @@ int ParseResponse(PyObject* response, int* isError, int* code, char** dataOrMess
...
 #else
-#define DUMP_STAT(x)
+# define DUMP_STAT(x) do {/* no-op */} while (0)
```
detectivelyw: What's this?

src/shared/platform/lind_platform.c
alyptik: Yeah, the macro just obfuscates this part, as it's really only used in one place. Just going to remove it and use a string literal.

src/shared/platform/lind_platform.c
alyptik: After some examination, it turns this macro isn't even useful, as it's only used in an `#if 0` section.  Removing the dead code.

src/shared/platform/lind_platform.c
alyptik: See previous resolution.

src/trusted/service_runtime/lind_syscalls.c
alyptik: Yes, agreed. This format is probably best, as it's already established in many places.

src/trusted/service_runtime/lind_syscalls.c
alyptik: This has been removed as per @JustinCappos in my other branches, but it does seem better to just not add it at all, yeah. Going to just rip it out (of course with testing to be sure this doesn't break anything).

src/shared/platform/lind_platform.c
alyptik: See previous resolution.

alyptik:
@detectivelyw @JustinCappos So thinking about this more, most of the problems that we've been finding are largely because at the start of all this I based most of the work off of caging and ended up just copying the Repy parts (in the beginning nothing seemed to work if I based my branches off of lind).

Now that it is starting to come together, mostly due to explanations of Repyv2 from @aaaaalbert, there seems to be a large amount of bitrot leftover from the original SVN imports, which seem to be causing a lot of code style issues. Things like old commented out code, unneeded macros, and oddly named variables are scattered about these sections.

So at this point I am wondering if you guys think it would be useful to rebuild the Repy and SeattleLib parts cleanly from scratch?

Since these issues are only style-related, as everything does work, I'm not sure if a major cleanup like this is useful to think about now, or maybe if we should just shelf it for later?

detectivelyw:
> @detectivelyw @JustinCappos So thinking about this more, most of the problems that we've been finding are largely because at the start of all this I based most of the work off of caging and ended up just copying the Repy parts (in the beginning nothing seemed to work if I based my branches off of lind).
> 
> Now that it is starting to come together, mostly due to explanations of Repyv2 from @aaaaalbert, there seems to be a large amount of bitrot leftover from the original SVN imports, which seem to be causing a lot of code style issues. Things like old commented out code, unneeded macros, and oddly named variables are scattered about these sections.
> 
> So at this point I am wondering if you guys think it would be useful to rebuild the Repy and SeattleLib parts cleanly from scratch?
> 
> Since these issues are only style-related, as everything does work, I'm not sure if a major cleanup like this is useful to think about now, or maybe if we should just shelf it for later?

I vote for shelving a major cleanup for later. 
Our codebase works now, and I think most code has good / consistent style. So I don't think it is necessary to rebuild from scratch. And I think fixing style-related issues as we keep building on the existing codebase is a reasonable and practical approach.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -32,7 +253,7 @@ void WINAPI NaClAppThreadLauncher(void *state) {
 
   NaClSignalStackRegister(natp->signal_stack);
 
-  NaClLog(4, "      natp = 0x%016"NACL_PRIxPTR"\n", (uintptr_t) natp);
+  NaClLog(4, "     natp  = 0x%016"NACL_PRIxPTR"\n", (uintptr_t)natp);
   NaClLog(4, " prog_ctr  = 0x%016"NACL_PRIxNACL_REG"\n", natp->user.prog_ctr);
   NaClLog(4, "stack_ptr  = 0x%016"NACL_PRIxPTR"\n",
```
detectivelyw, line 253: Is there any reason for us to use "4" here in the NaClLog? Other messages are directed to the standard output "1".

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -147,13 +426,12 @@ void NaClAppThreadTeardown(struct NaClAppThread *natp) {
   NaClLog(3, " freeing thread object\n");
   NaClAppThreadDelete(natp);
   NaClLog(3, " NaClThreadExit\n");
+
   NaClThreadExit();
-  NaClLog(LOG_FATAL,
-          "NaClAppThreadTeardown: NaClThreadExit() should not return\n");
+  NaClLog(LOG_FATAL, "NaClAppThreadTeardown: NaClThreadExit() should not return\n");
```
detectivelyw, line 422: So here we use "LOG_FATAL", but at other places we were using numbers like 1, 2, 3, 4. I wonder if we could make it consistent and easy to read / understand.

alyptik: It's a bit odd because the number refers to number of -v needed to see the message, with macros for the special cases. I am aware it's not ideal, and I'm still trying to figure out of there's a better way. This would definitely be a change for a seperate PR, though.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: This particular log message is fairly noisy and not usually useful information, so I have made it require -vvvv to be visible.

alyptik:
@detectivelyw Yeah, I am sort of leaning that way as well. Also interested in your thoughts, @aaaaalbert and @moyix

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -81,25 +81,20 @@ struct NaClApp *NaClChildNapCtor(struct NaClAppThread *natp) {
   NaClXMutexUnlock(&nap_master->children_mu);
 
   /* store cage_ids in both master and parent to provide redundancy and avoid orphans */
-  for (struct NaClApp *nap_cur = nap_master; nap_cur; nap_cur = nap_parent) {
-    NaClXMutexLock(&nap_cur->children_mu);
-    if (nap_cur->children_ids[nap_cur->num_children]) {
-       nap_cur->num_children++;
+  for (struct NaClApp *cur = nap_master; cur == nap_master || cur == nap_parent; cur = nap_parent) {
```
detectivelyw: In my opinion, nap_cur seems like a better name than cur, since it tells us more about what data structure it represents.

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -57,8 +59,28 @@
 #include "native_client/src/trusted/simple_service/nacl_simple_service.h"
 #include "native_client/src/trusted/threading/nacl_thread_interface.h"
 
+double time_counter = 0.0;
+double time_start = 0.0;
+double time_end = 0.0;
+int cage = 1001;
```
detectivelyw: I believe this (cage = 1001) isn't used any more. Please help remove it.

aaaaalbert:
So, @detectivelyw and @alyptik, you suggest:
> I vote for shelving a major cleanup for later.

I don't think this would be a major cleanup actually. As far as I can see, none of the difficult things you are trying to achieve with NaCl/Repy are more difficult with an up-to-date version of repy_v2 / seattlelib_v2. IMHO you'd get lots of benefits from updating.

* The build process isn't based off `preparetest.py` anymore, see [the new build docs](https://github.com/SeattleTestbed/docs/blob/master/Contributing/BuildInstructions.md).
* Through this, the codebase is laid out in a more meaningful way. We've split the subdirs of our old SVN (i.e. what you have here!) into indepedent repos.
* This makes it more straightforward to select and use (and modify) only the parts that you are interested in, e.g. *not* the nodemanager or experimenter shell.
* The preprocessing step (from `.mix` to `.repy`) is replaced by a more Python-like import process using `dylink`.

That's what I can think off right now, perhaps there are other things that I don't see yet where it matters more.

src/trusted/service_runtime/nacl_app_thread.c
alyptik: Yeah you are right, will revert the rename part here.

src/trusted/service_runtime/sel_ldr.c
alyptik: Indeed it's not, will remove.

alyptik:
@aaaaalbert After talking with @JustinCappos it actually seems most of the code is v2, just not the most recent version of it. It shouldn't be too much work to clean up and will be giving it a go in a test branch, thanks for the feedback on this.

alyptik:
The feedback was all very helpful guys, thanks! I've addressed what was feasible and am merging it in.

aaaaalbert:
> After talking with @JustinCappos it actually seems most of the code is v2, just not the most recent version of it.

Exactly, the single biggest difference is in the dir/repo layout and build scripts.

PR #7, opened 2018-10-06
Add internal IRT implementation of execv()/execve()

alyptik:
Add `execv()` and `execve()` system call implementations to the internal runtime.

Implementation Details:
- Copy the argument and environment vectors from the "parent" NaClApp.
- Duplicate its `cage_id` along with any other relevant internal state.
- Leverage the existing fork() implementation to emulate replacement of the entire process image.

### DIFF
--- a/src/trusted/service_runtime/nacl_globals.h
+++ b/src/trusted/service_runtime/nacl_globals.h
@@ -129,18 +129,13 @@ void  NaClGlobalModuleFini(void);
 /* this is defined in src/trusted/service_runtime/arch/<arch>/ sel_rt.h */
 void NaClInitGlobals(void);
 
-static INLINE void NaClPatchAddr(uintptr_t child_bits,
-                                 uintptr_t parent_bits,
-                                 uintptr_t *start,
-                                 size_t cnt) {
+static INLINE void NaClPatchAddr(uintptr_t child_bits, uintptr_t parent_bits, uintptr_t *start, size_t cnt) {
   for (size_t i = 0; i < cnt; i++) {
-    /* skip addresses with matching high bits */
-    if (!((parent_bits ^ start[i]) >> NACL_PAGESHIFT)) {
+    if ((parent_bits >> NACL_PAGESHIFT) != (start[i] >> NACL_PAGESHIFT))
       continue;
-    }
-    NaClLog(2, "patching %p\n", (void *)start[i]);
+    NaClLog(1, "patching %p\n", (void *)start[i]);
     start[i] = child_bits | (start[i] & UNTRUSTED_ADDR_MASK);
-    NaClLog(2, "new addr %p\n", (void *)start[i]);
+    NaClLog(1, "new addr %p\n", (void *)start[i]);
   }
 }
 
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -856,18 +856,24 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
            "%"NACL_PRIdS"[0x%"NACL_PRIxS"])\n",
           (uintptr_t) natp, d, (uintptr_t) buf, count, count);
 
+  /* check for closed fds */
+  if (fd < 0) {
+    retval = -NACL_ABI_EBADF;
+    goto out;
+  }
+
   ndp = NaClGetDesc(nap, fd);
   NaClLog(2, " ndp = %"NACL_PRIxPTR"\n", (uintptr_t) ndp);
   if (!ndp) {
     retval = -NACL_ABI_EBADF;
-    goto cleanup;
+    goto out;
   }
 
   sysaddr = NaClUserToSysAddrRange(nap, (uintptr_t) buf, count);
   if (kNaClBadAddress == sysaddr) {
     NaClDescUnref(ndp);
     retval = -NACL_ABI_EFAULT;
-    goto cleanup;
+    goto out;
   }
 
   /*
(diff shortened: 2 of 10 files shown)


src/trusted/service_runtime/nacl_globals.h
```diff
@@ -24,6 +24,7 @@ EXTERN_C_BEGIN
 #define PROT_RW (NACL_ABI_PROT_READ | NACL_ABI_PROT_WRITE)
 #define PROT_RX (NACL_ABI_PROT_READ | NACL_ABI_PROT_EXEC)
 #define F_ANON_PRIV (NACL_ABI_MAP_PRIVATE | NACL_ABI_MAP_ANONYMOUS)
+#define SHM_SIZE (1u << 13)
```
JustinCappos: why?  Also, will this always be a constant?

src/trusted/service_runtime/sel_ldr_standard.c
```diff
@@ -727,11 +725,11 @@ int NaClCreateMainThread(struct NaClApp     *nap,
    * data.  We are assuming that the caller is non-adversarial and the
    * code does not look like string data....
    */
-  for (i = 0; i < argc && argv; ++i) {
+  for (i = 0; i < argc && argv && argv[i]; ++i) {
```
JustinCappos: Why ++i instead of i++?

src/trusted/service_runtime/nacl_globals.h
alyptik: Whoops, that was leftover code from something else I had planned but decided wasn't worth pursuing! Good catch, removing.

src/trusted/service_runtime/sel_ldr_standard.c
alyptik: The rest of the `for()` loops in this file use `++i`, I figured it'd be good to stay consistent.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -3965,38 +3965,32 @@ int32_t NaClSysPipe2(struct NaClAppThread *natp, uint32_t *pipedes, int flags) {
...
+  nap_child->running = 0;
   ret = nap_child->cage_id;
-  if (!NaClCreateMainForkThread(nap, natp, nap_child, child_argc, child_argv, NULL)) {
+
+  /* start fork thread */
+  if (!NaClCreateMainForkThread(nap, natp, nap_child, child_argc, child_argv, nap_child->clean_environ)) {
     NaClLog(1, "%s\n", "[NaClSysFork] forking program failed!");
-    ret = -NACL_ABI_ENOMEM;
-    goto out;
+    ret = -NACL_ABI_ENOEXEC;
```
moyix: Why the change to ENOEXEC here? I don't think that's a standard error for `fork` (it is for exec, e.g. in case the thing you're trying to exec isn't a properly formatted executable).

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -456,62 +456,88 @@ int32_t NaClSysNameService(struct NaClAppThread *natp,
...
-  fd_cage_table[nap->cage_id][newfd] = fd_cage_table[nap->cage_id][oldfd];
-  ret = newfd;
-  nap->fd++;
+  old_desc = fd_cage_table[nap->cage_id][oldfd];
+  if (!(old_nd = NaClGetDesc(nap, old_desc))) {
+    ret= -NACL_ABI_EBADF;
+    goto out;
+  }
+  new_desc = NaClSetAvail(nap, old_nd);
+  fd_cage_table[nap->cage_id][newfd] = new_desc;
```
moyix, line 529: Does `newfd` need to first be closed if it's open (per the `dup2` man page)? Or is that done by `NaClSetAvail` somehow?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -456,62 +456,88 @@ int32_t NaClSysNameService(struct NaClAppThread *natp,
...
 
-done:
+out:
   return ret;
 }
 
 int32_t NaClSysDup3(struct NaClAppThread  *natp,
                     int                   oldfd,
                     int                   newfd,
                     int                   flags) {
```
moyix, line 540: As a general point, can `dup` and `dup2` be implemented as calls to `dup3` with the appropriate flags?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -2792,7 +2790,7 @@ int32_t NaClSysImcRecvmsg(struct NaClAppThread         *natp,
   recv_hdr.iov_length = kern_nanimh.iov_length;
 
   recv_hdr.ndescv = new_desc;
-  recv_hdr.ndesc_length = NACL_ARRAY_SIZE(new_desc);
+  recv_hdr.ndesc_length = sizeof new_desc / sizeof *new_desc;
```
moyix: Would prefer to always use parentheses with sizeof – makes things a bit easier to read and removes ambiguity.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -2881,7 +2879,7 @@ int32_t NaClSysImcRecvmsg(struct NaClAppThread         *natp,
 cleanup:
   /* copy out updated desc count, flags */
   if (retval < 0) {
-    for (i = 0; i < NACL_ARRAY_SIZE(new_desc); ++i) {
+    for (i = 0; i < sizeof new_desc / sizeof *new_desc; ++i) {
```
moyix: Parentheses again

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: After checking the manpage and POSIX standard, it seems you are right and `ENOEXEC` isn't even a valid `errno`. Changing it back to `ENOMEM`.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Hm, well `dup()` not really, since it returns an arbitrary file descriptor, but I think it may be possible with `dup2()`/`dup3`. I'll see if I can rework `dup3()` to just call `dup2()`.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Good catch, I completely forget to close `newfd`. Fixing.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Sure, I can do that.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Will fix.

src/trusted/service_runtime/nacl_syscall_common.c
alyptik: Duplicated code removed.

alyptik:
Thank you for the quick review, merged.

Issue #1, opened 2018-06-30
Finalize Travis CI Integration


alyptik:
The current goal is to have this repository auto-trigger builds at the main Travis CI instance [[1]](https://travis-ci.org/Lind-Project/lind_project/settings) when commits are pushed.

[1] [https://travis-ci.org/Lind-Project/lind_project/settings](https://travis-ci.org/Lind-Project/lind_project/settings)

alyptik:
Webhooks should be triggering automated builds on pushes now.

alyptik:
This is still something that should be revisited at some point, as a working CI would be very useful to have.

lukpueh:
I created [this little Flask webapp](https://gist.github.com/lukpueh/498ff3489321bdc7106c05a2fd5b941c) that listens for push notifications on the `native_client` repo (1) originating from a [GitHub webhook](https://developer.github.com/webhooks/) and makes a call to the [Travis API](https://docs.travis-ci.com/user/triggering-builds/) to  trigger a build on the `lind_project` repo (2).

Feel free to adopt and use it as you see fit. Alternatively, I can set it up for you, if you give me admin access to the Lind project organization. :)

One thing I'm not so sure about is how (1) and (2) are related. I see that you have a bunch of submodules in (2), but none of them point to (1), right?

So in order for this webhook to make sense, a change in (1) should somehow propagate to (2).

alyptik:
Sorry was a bit busy over the weekend, sent you the invite @lukpueh, thanks!

lukpueh:
Thanks, @alyptik. Before I set anything up though, we should discuss the questions I have. Either here and/or later today during the dev meeting.

alyptik:
@lukpueh Hope that kind of explained things a bit, if you have any other questions feel free to email me

lukpueh:
Okay, so from our discussion on Tuesday I gathered that, when building `Lind-project/lind_project` it git clones `Lind-project/native_client` at branch `develop`  (see [mklind#L190](https://github.com/Lind-Project/lind_project/blob/master/mklind#L190) and [mklind#L296-L297](https://github.com/Lind-Project/lind_project/blob/master/mklind#L296-L297)).

In order for the proposed trigger-push-via-webhook setup to be effective the  hardcoded `develop` branch in [mklind#L190](https://github.com/Lind-Project/lind_project/blob/master/mklind#L190) would need to be changed to the commit id being pushed before building.

[...]

PR #8, opened 2018-11-04
Add working Travis CI configuration

alyptik:
I decided to just split out the glibc parts of the build which simplified everything significantly. As an added bonus, with this particular solution we wouldn't need to worry about hosting or maintaining webhooks anymore.

@lukpueh, since you have the most experience with this stuff, does this approach seem alright to you? It isn't the most elegant solution, but this seems to be the most reliable way to test NaCl changes without having to rebuild the entire toolchain every time we push a commit: [native_client/builds/450379081](https://travis-ci.org/Lind-Project/native_client/builds/450379081).

After some thought, I realized we barely ever change glibc itself, so it's pretty wasteful to rebuild it every time we push a commit to NaCl. Since nothing in NaCl would be able to affect the success of our glibc build, it's been completely divorced from this repository.

The script commands are used to ignore the non-zero exit codes, while still making sure the commands themselves start up correctly and actually load our binaries. They will, however, always fail after ELF loading, since there is currently no way to tell `docker build` to mount `/dev/shm` executable (`docker run` provides the `--ipc=host` flag for this) in just the `Dockerfile`. NaCl requires executable `/dev/shm` because it maps our ELF there and executes that memory at runtime.

We could possibly work around this using `docker compose` later on, but that is a bit more complicated and should probably be handled in another PR.

The CI for glibc is still a bit problematic, as it times out even when building nothing **except** glibc: [Dockerfile#L24](https://github.com/Lind-Project/lind_project/blob/develop/docker/prebuiltsdk/Dockerfile#L24). Since it's not something we change very often, I don't believe it should be a high priority right now, and is best handled later with a PR to the glibc subrepository itself.

For reference, the main repository Travis CI builds can be found [here](https://travis-ci.org/Lind-Project/lind_project/builds).

Signed-off-by: Joey Pabalinas <joeypabalinas@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,14 +17,15 @@ matrix:
             - g++-multilib
 
 script:
-    # special case pull requests so they build the source and not the target branch
-    - if [[ "$TRAVIS_PULL_REQUEST" == false ]]; then cur_branch="$TRAVIS_BRANCH"; else cur_branch="$TRAVIS_PULL_REQUEST_BRANCH"; fi
-    # skip building the master and lind branches as those don't exist in all subrepositories
-    #
-    # TODO: make subrepository branch names consistent and remove special casing of master and lind -jp
-    - if [[ "$cur_branch" == master || "$cur_branch" == lind ]]; then cur_branch="develop"; fi
-    # run a test of the loader after building
-    - docker build --build-arg=BRANCH="$cur_branch" -t alyptik/lind .
-    - docker run --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind lind /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched; exec /fork"'
+  # special case pull requests so they build the source and not the target branch
+  - if [[ "$TRAVIS_PULL_REQUEST" == false ]]; then cur_branch="$TRAVIS_BRANCH"; else cur_branch="$TRAVIS_PULL_REQUEST_BRANCH"; fi
+  # skip building the master and lind branches as those don't exist in all subrepositories
+  #
+  # TODO: make subrepository branch names consistent and remove special casing of master and lind -jp
+  - if [[ "$cur_branch" == master || "$cur_branch" == lind ]]; then cur_branch="develop"; fi
+  # travis_wait needed due to build log exceeding Travis CI's maximum length
+  - travis_wait 60 bash -c "docker build --build-arg=BRANCH=$cur_branch -t alyptik/lind . >/dev/null 2>&1"
+  # run a test of the loader after building
+  - docker run --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind lind /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched; exec /fork"'
 
 # vi:ft=yaml:
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ script:
     # be examined manually.
     #
     # TODO: modify loader to only return non-zero on errors -jp
-    - docker build -t alyptik/lind:naclruntime .
+    - docker build --build-arg=TRAVIS_COMMIT="$TRAVIS_COMMIT" -t alyptik/lind:naclruntime .
     - docker run --rm --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind:naclruntime "$REPY_PATH/bin/lind" /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched"' || true
 
 # vi:ft=yaml:
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,8 @@ script:
     # be examined manually.
     #
     # TODO: modify loader to only return non-zero on errors -jp
-    - docker build --build-arg=TRAVIS_COMMIT="$TRAVIS_COMMIT" -t alyptik/lind:naclruntime .
+    - if [[ "$TRAVIS_PULL_REQUEST" == false ]]; then BRANCH="$TRAVIS_BRANCH"; else BRANCH="$TRAVIS_PULL_REQUEST_BRANCH"; fi
+    - docker build --build-arg=BRANCH="$BRANCH" -t alyptik/lind:naclruntime .
     - docker run --rm --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind:naclruntime "$REPY_PATH/bin/lind" /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched"' || true
 
 # vi:ft=yaml:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 dist: trusty
 group: deprecated-2017Q4
 sudo: required
-env:
-  - REPY_PATH="/home/lind/lind_project/lind/repy"
 services:
   - docker
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,18 @@ dist: trusty
 language: c
 group: deprecated-2017Q4
 sudo: required
+
+env:
+  - REPY_PATH="/home/lind/lind_project/lind/repy"
+
 services:
   - docker
+
 git:
     depth: 1
 
 matrix:
   include:
-    # works on Precise and Trusty
     - os: linux
       addons:
         apt:
@@ -21,7 +25,15 @@ matrix:
             - g++-multilib
 
 script:
+    # Run a test of the loader after building. Since the loader always
+    # returns non-zerom even if there were no errors, the results must
+    # be examined manually.
+    #
+    # TODO: modify loader to only return non-zero on errors -jp
     - docker build -t alyptik/lind:naclruntime .
-    - docker run --rm --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind:naclruntime /home/lind/lind_project/lind/repy/bin/lind /bin/bash -c 'a=potato; [[ $a =~ pota.* ]] && echo foo; echo bar' || true
+    - docker run --rm --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind:naclruntime \
+      /home/lind/lind_project/lind/repy/bin/lind \
+      /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched"' \
+      || true
 
 # vi:ft=yaml:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,10 @@ dist: trusty
 language: c
 group: deprecated-2017Q4
 sudo: required
-
+git:
+  submodules: false
 env:
   - REPY_PATH="/home/lind/lind_project/lind/repy"
-
 services:
   - docker
 
@@ -22,13 +22,14 @@ matrix:
             - g++-multilib
 
 script:
-    # Run a test of the loader after building. Since the loader always
-    # returns non-zerom even if there were no errors, the results must
-    # be examined manually.
+    # special case pull requests so they build the source and not the target branch
+    - if [[ "$TRAVIS_PULL_REQUEST" == false ]]; then cur_branch="$TRAVIS_BRANCH"; else cur_branch="$TRAVIS_PULL_REQUEST_BRANCH"; fi
+    # skip building the master and lind branches as those don't exist in all subrepositories
     #
-    # TODO: modify loader to only return non-zero on errors -jp
-    - if [[ "$TRAVIS_PULL_REQUEST" == false ]]; then BRANCH="$TRAVIS_BRANCH"; else BRANCH="$TRAVIS_PULL_REQUEST_BRANCH"; fi
-    - docker build --build-arg=BRANCH="$BRANCH" -t alyptik/lind:naclruntime .
-    - docker run --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind:naclruntime "$REPY_PATH/bin/lind" /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched"' || true
+    # TODO: make subrepository branch names consistent and remove special casing of master and lind -jp
+    - if [[ "$cur_branch" == master || "$cur_branch" == lind ]]; then cur_branch="develop"; fi
+    # run a test of the loader after building
+    - docker build --build-arg=BRANCH="$cur_branch" -t alyptik/lind .
+    - docker run --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind "$REPY_PATH/bin/lind" /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched; exec /fork"'
 
 # vi:ft=yaml:
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,9 +31,6 @@ script:
     #
     # TODO: modify loader to only return non-zero on errors -jp
     - docker build -t alyptik/lind:naclruntime .
-    - docker run --rm --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind:naclruntime \
-      /home/lind/lind_project/lind/repy/bin/lind \
-      /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched"' \
-      || true
+    - docker run --rm --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind:naclruntime "$REPY_PATH/bin/lind" /bin/bash -c 'a="regex_test_string"; [[ $a =~ reg.*ng$ ]] && echo "string matched"' || true
 
 # vi:ft=yaml:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+dist: trusty
+language: c
+group: deprecated-2017Q4
+sudo: required
+services:
+  - docker
+git:
+    depth: 1
+git:
+  submodules: false
+
+matrix:
+  include:
+    # works on Precise and Trusty
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - build-essential
+            - gcc-multilib
+            - g++-multilib
+
+before_install:
+    - git submodule update --init --recursive --remote .
+
+script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,32 @@
+dist: trusty
+language: c
+group: deprecated-2017Q4
+sudo: required
+services:
+  - docker
+git:
+    depth: 1
+git:
+  submodules: false
```
lukpueh: Why disable travis automatic submodule handling? This repos does not have any submodules.

.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - build-essential
+            - gcc-multilib
+            - g++-multilib
+
+before_install:
+    - git submodule update --init --recursive --remote .
```
lukpueh: Why explicitly updating submodules? This repo does not have any submodules.

.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+
+script:
+    - docker build -t alyptik/lind:naclruntime .
+    - docker run --rm --ipc=host --cap-add=SYS_PTRACE -it alyptik/lind:naclruntime /home/lind/lind_project/lind/repy/bin/lind /bin/bash -c 'a=potato; [[ $a =~ pota.* ]] && echo foo; echo bar' || true
+
```
lukpueh, line 30: Do I understand correctly that the docker run command will execute:
```shell
/usr/bin/zsh  /home/lind/lind_project/lind/repy/bin/lind /bin/bash -c 'a=potato; [[ $a =~ pota.* ]] && echo foo; echo bar' || true
```
?
Do you mind adding a comment what this command (chain) does?

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN ["/usr/bin/env", "lindfs", "cp", "./hello"]
+RUN ["/usr/bin/env", "lindfs", "cp", "./.bashrc"]
+RUN ["/usr/bin/env", "lindfs", "cp", "./.bash_logout"]
+RUN ["/usr/bin/env", "lindfs", "cp", "./.bash_profile"]
+RUN ["/usr/bin/env", "script", "-qfc", "$(printf '%q ' lind /fork)", "/dev/null"]
+RUN ["/usr/bin/env", "script", "-qfc", "$(printf '%q ' lind /hello)", "/dev/null"]
+RUN ["/usr/bin/env", "script", "-qfc", "$(printf '%q ' lind /bin/bash --version)", "/dev/null"]
+
+ENTRYPOINT ["/usr/bin/zsh"]
+CMD ["-c", "lind /fork; lind /hello; lind /bin/bash --version; exec zsh"]
```
lukpueh: The arguments to `ENTRYPOINT`, specified in `CMD` are overridden by the arguments passed to `docker run` in `.travis.yml`.
I suppose above arguments are for testing and these here for general container usage? I think this should be documented somewhere.

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+ENV PATH "$NACL_SDK_ROOT/toolchain/linux_x86_glibc/bin:$PATH"
+ENV PATH "$REPY_PATH/bin:$PATH"
+ENV PATH "$LIND_BASE:$PATH"
+ENV PATH "/root/bin:/home/lind/bin:$PATH"
+ENV PATH "/root/.local/bin:/home/lind/.local/bin:$PATH"
+
+USER lind
+
+WORKDIR /home/lind/lind_project/
+RUN ["./mklind", "-qe", "nacl"]
```
lukpueh: So this is where you pull in the submodule of `native_client` into the base container, right? Are you sure this is happening? The [travis log](https://travis-ci.org/Lind-Project/native_client/builds/450379081#L796-L827) does not look like it. Maybe the docker image `alyptik/lind:prebuiltsdk` is not up to date?

Also, as pointed out in https://github.com/Lind-Project/native_client/issues/1#issuecomment-435068998. If you want to trigger a build that pulls in changes that you pushed to any branch on `native_client` you need to instruct `mklind` to update the submodule to that commit/branch. Currently, you're always updating to the tip of the `develop` branch (see [`git submodule --remote...`](https://git-scm.com/docs/git-submodule/2.0.0#git-submodule---remote) docs and your [`.gitmodules`](https://github.com/Lind-Project/lind_project/blob/master/.gitmodules#L38) `lind_project` branch config for `native_client`).

Maybe you can pass on the branch name from travis environment variables to `mklind` (see [travis docs](https://docs.travis-ci.com/user/environment-variables#convenience-variables) and this [related blog post](https://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/))?

lukpueh: Not directly related to this PR, but is the [`git clone` fallback in `mklind`](https://github.com/Lind-Project/lind_project/blob/master/mklind#L238-L244) necessary? And is it really only fallback, or does it also interfere with the `git submodule update` call above?

lukpueh (commented):
I am not familiar enough with the lind build process to know whether the build sequence in you dockerfile is accurate. However, I did add some general comments and questions regarding git submodules. Let me know if anything is unclear.

lukpueh:
@alyptik This approach definitely seems better than using webhooks. :)

.travis.yml
alyptik: Good catch, removing.

.travis.yml
alyptik: Same.

.travis.yml
alyptik: Sure, will add.

Dockerfile
alyptik: Yeah, I'm actually going to change this to CMD to allow it to be overridden. Not quite sure where to put the comment though, the `README` maybe?

Dockerfile
alyptik: This is actually literally only building `nacl` and `repy` in this `Dockerfile`. I split it up into three, the first downloads and checks out all the submodules, the second one builds glibc (this one doesn't pass Travis CI yet), and this one just does the final bits of the whole process. Even with only these steps I am very close to the 50min limit of Travis CI, so I can't really see any other way to do this.

alyptik:
@lukpueh The base `Dockerfile` which does the cloning and checking out is... not the nicest I've ever written: [lind_project/docker/base/Dockerfile](https://github.com/Lind-Project/lind_project/blob/master/docker/base/Dockerfile). Most of that was for ease of use when testing everything, and it definitely needs to be trimmed, but it does work for now at least.

Dockerfile
lukpueh: For starters you could add a [comment to the `Dockerfile`](https://docs.docker.com/engine/reference/builder/#format). Eventually, documentation about the whole build process in the `README.md` would be nice.

Dockerfile
alyptik: Ah, I should probably remove that `# fallback` comment, as it's not really a fallback at all; it's there to set up the submodule symlinks correctly.

Dockerfile
alyptik: That sounds good to me, will take care of that.

lukpueh:
@alyptik Let me know when you want me to give this PR another review pass.

alyptik:
@lukpueh Sure, will let you know, thanks.

alyptik:
@lukpueh I think it's good to go for another review pass! Main changes are the passing of the current branch as a build argument which is used in places like [Lind-Project/lind_project/mklind#L256](https://github.com/Lind-Project/lind_project/blob/master/mklind#L256).

lukpueh:
Alright, let me try to see if I understand correctly what's happening here, by going through the [`native_client/Dockerfile`](https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile) which is used for the travis build ([`native_client/.travis.yml#L27`](https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/.travis.yml#L27)):

https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L1
The build is based on the `alyptik/lind:prebuiltsdk` image on docker hub.  We don't know much about that image here,  but I assume it was built using [`lind_project/docker/prebuiltsdk/Dockerfile`](https://github.com/Lind-Project/lind_project/blob/master/docker/base/Dockerfile). If that's correct, it itself is based on `alyptik/lind:base`, probably built using [`lind_project/docker/base/Dockerfile`](https://github.com/Lind-Project/lind_project/blob/master/docker/base/Dockerfile).
Looking at the base dockerfile, we know that we have some version of `lind_project/mklind` and have already used it to call `mklind download`, which sets up the base repo (i.e. `lind_project`) and it's submodules.

So far so good (?) ... 

https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L6
Next in the dockerfile we read the docker build variable `BRANCH` which came from travis, telling us on which branch the travis `native_client` build was triggered.

https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L16
Then we change into the `lind_project` repository that was put there in `alyptik/lind:base`.
We don't really know here what branch it is on and at what commit...

https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L17
... but we are updating the current branch (according to the travis log it is `master`) from the remote that was defined in `alyptik/lind:base` (according to the travis log it is https://github.com/Lind-Project/lind_project) and we fetch all tags and associated objects (`-t`) and submodules in parallel (`-j8`).

https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L18
Then we update the branches of all submodules as defined in [`lind_project/.gitmodules`](https://github.com/Lind-Project/lind_project/blob/master/.gitmodules#L38). In case of the `native_client` submodule that is `develop`.

https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L19-L20
Then we use `mklind` for targets `nacl` (submodule `Lind-Project/native_client`), `repy` (submodule `Lind-Project/nacl_repy`) and `install`.
*NOTE: I suppose doing this twice is not voluntary, and neither is the mixed case in the docker directive `RUn`?*

Now we have to go back to `lind_project/mklind` to see what's happening. In [`mkdlind:L467-L470`](https://github.com/Lind-Project/lind_project/blob/master/mklind#L467-L470) we fetch all remotes and check out the submodule to the branch that triggered the travis build, as propagated by `BRANCH`. 
*NOTE: Why do you always return 0 on checkout? (see [`git checkout "$BRANCH" || true`](https://github.com/Lind-Project/lind_project/blob/master/mklind#L469))

So we finally have to right version of `native_client`, i.e. the one that triggered the travis build, and can build it inside `lind_project`. Hooray! :) 

If we also take a look at the function called by `mklind repy` ([mklind#L326-L329](https://github.com/Lind-Project/lind_project/blob/master/mklind#L326-L329)) we see that it also tries to checkout the `Lind-Project/nacl_repy` submodule at `BRANCH`. How do we now if that repo has that branch? And why would we want that branch?


**Bottom line**
What I'm trying to say here is, the whole process seems mighty convoluted and makes it hard to retrace which version of what is used when and how. I think this is what @aaaaalbert warned about.
Unfortunately, I can't really give good suggestions on how to make the Lind build process easier, and I still think that passing through the BRANCH name from the travis environment is okay.

However, I really think that, especially in the `Dockerfile`s and also in `mklind`, we could use a lot more comments and maybe also some print statements that make it clear what versions (by commit id) of files/repos are used.

alyptik:
@lukpueh wrote:
> The build is based on the `alyptik/lind:prebuiltsdk` image on docker hub. We don't know much about that image here, but I assume it was built using [`lind_project/docker/prebuiltsdk/Dockerfile`](https://github.com/Lind-Project/lind_project/blob/master/docker/base/Dockerfile). If that's correct, it itself is based on `alyptik/lind:base`, probably built using [`lind_project/docker/base/Dockerfile`](https://github.com/Lind-Project/lind_project/blob/master/docker/base/Dockerfile).
> Looking at the base dockerfile, we know that we have some version of `lind_project/mklind` and have already used it to call `mklind download`, which sets up the base repo (i.e. `lind_project`) and it's submodules.
> So far so good (?) ...

Correct.

Basically the main steps in the build are:

* base:
  * ./mklind download
* prebuiltsdk:
  * ./mklind glibc
* naclruntime:
  * ./mklind nacl
  * ./mklind repy
  * ./mklind install

which correspond to the three `Dockerfile`s.

> [native_client/Dockerfile](https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L6)
> 
> Line 6 in [998c2a7](/Lind-Project/native_client/commit/998c2a7e726715bcc0583648500d5661aa59dbd1)
> 
>  ARG BRANCH 
> 
> Next in the dockerfile we read the docker build variable `BRANCH` which came from travis, telling us on which branch the travis `native_client` build was triggered.
> [native_client/Dockerfile](https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L16)
> 
> Line 16 in [998c2a7](/Lind-Project/native_client/commit/998c2a7e726715bcc0583648500d5661aa59dbd1)
> 
>  WORKDIR /home/lind/lind_project/ 
> 
> Then we change into the `lind_project` repository that was put there in `alyptik/lind:base`.
> We don't really know here what branch it is on and at what commit...
> [native_client/Dockerfile](https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L17)
> 
> Line 17 in [998c2a7](/Lind-Project/native_client/commit/998c2a7e726715bcc0583648500d5661aa59dbd1)
> 
>  RUN git pull -t -j8 
> 
> ... but we are updating the current branch (according to the travis log it is `master`) from the remote that was defined in `alyptik/lind:base` (according to the travis log it is https://github.com/Lind-Project/lind_project) and we fetch all tags and associated objects (`-t`) and submodules in parallel (`-j8`).

Indeed.

> [native_client/Dockerfile](https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L18)
> 
> Line 18 in [998c2a7](/Lind-Project/native_client/commit/998c2a7e726715bcc0583648500d5661aa59dbd1)
> 
>  RUN git submodule update --recursive --remote . 
> 
> Then we update the branches of all submodules as defined in [`lind_project/.gitmodules`](https://github.com/Lind-Project/lind_project/blob/master/.gitmodules#L38). In case of the `native_client` submodule that is `develop`.
> [native_client/Dockerfile](https://github.com/Lind-Project/native_client/blob/998c2a7e726715bcc0583648500d5661aa59dbd1/Dockerfile#L19-L20)

I realize now this line might be redundant.

> Lines 19 to 20 in [998c2a7](/Lind-Project/native_client/commit/998c2a7e726715bcc0583648500d5661aa59dbd1)
> 
>  RUN for t in nacl repy install; do ./mklind -q "$t"; done 
> 
> Then we use `mklind` for targets `nacl` (submodule `Lind-Project/native_client`), `repy` (submodule `Lind-Project/nacl_repy`) and `install`.
> _NOTE: I suppose doing this twice is not voluntary, and neither is the mixed case in the docker directive `RUn`?

Yeah, fixing that now :)

> Now we have to go back to `lind_project/mklind` to see what's happening. In [`mkdlind:L467-L470`](https://github.com/Lind-Project/lind_project/blob/master/mklind#L467-L470) we fetch all remotes and check out the submodule to the branch that triggered the travis build, as propagated by `BRANCH`.
> *NOTE: Why do you always return 0 on checkout? (see [`git checkout "$BRANCH" || true`](https://github.com/Lind-Project/lind_project/blob/master/mklind#L469))

I should probably add a comment there; this is temporary until I make all the branch names consistent across our repositories. Some branches such as the `lind` branch in `native_client` don't have matching branches in the other submodules and so it would fail if you tried to, say, checkout the `lind` branch of `nacl_repy`.

> So we finally have to right version of `native_client`, i.e. the one that triggered the travis build, and can build it inside `lind_project`. Hooray! :)
> 
> If we also take a look at the function called by `mklind repy` ([mklind#L326-L329](https://github.com/Lind-Project/lind_project/blob/master/mklind#L326-L329)) we see that it also tries to checkout the `Lind-Project/nacl_repy` submodule at `BRANCH`. How do we now if that repo has that branch? And why would we want that branch?

Yeah, I'm still sort of trying to think of a clean way to handle that case where all we want to checkout is the Travis branch we are building.

> However, I really think that, especially in the `Dockerfile`s and also in `mklind`, we could use a lot more comments and maybe also some print statements that make it clear what versions (by commit id) of files/repos are used.

Agreed, will work on that.

> **Bottom line**
> What I'm trying to say here is, the whole process seems mighty convoluted and makes it hard to retrace which version of what is used when and how. I think this is what @aaaaalbert warned about.
> Unfortunately, I can't really give good suggestions on how to make the Lind build process easier, and I still think that passing through the BRANCH name from the travis environment is okay.

Haha, after trying to actually make it work myself, I can now see _exactly_ what @aaaaalbert was talking about. At this point do you think this is even worth bothering with? The amount of headache to make it work seems kind of crazy, and even after all of that the end result still seems very fragile to me...

Very helpful Round 2 review, appreciate it :)

lukpueh:
You're welcome, @alyptik! I think it is worth bothering, especially now that you've already bothered! :)

But as I said, comment comments comments. I spent about 2 hours for the review, and I suppose, if I looked at the same thing in a couple of weeks, I'd spend roughly the same amount of time again to understand what's going on.

So as long as you add enough documentation and a "this is a temporary solution until we have figured out how to build lind in a reasonable way"-disclaimer, it is okay.

alyptik:
@lukpueh said:
> You're welcome, @alyptik! I think it is worth bothering, especially now that you've already bothered! :)
> 
> But as I said, comment comments comments. I spent about 2 hours for the review, and I suppose, if I looked at the same thing in a couple of weeks, I'd spend roughly the same amount of time again to understand what's going on.
> 
> So as long as you add enough documentation and a "this is a temporary solution until we have figured out how to build lind in a reasonable way"-disclaimer, it is okay.

Yes, that's the main thing I'm worried about: to both myself coming back to this months later or a new person coming onto this cold it's pretty damn confusing, and I can't see any simple ways to dynamically build the correct branch without making this already convoluted code base even worse :( Just building `devel` and/or `master` isn't quite as flexible, but I'm starting to wonder if the simple approach is the best one here. 

I'll talk to the rest of the Lind team more at the meeting today and see what their opinions on this are, as well.

PR #11, opened 2019-09-03
Add Caging/Fork RPC to NaCl

rennergade:
Per: [Lind Issue 10](https://github.com/Lind-Project/lind_project/issues/10)

Adds cageid transmission to syscalls that utilize and require file descriptors.

We use the cageid, which is stored in each process's `nap` struct, and pass that to any new file descriptor when it is opened. From there it's passed as an extra parameter in any lind RPC call (ie `lind_open()` or `lind_write()`.

PR across several submodules:
https://github.com/Lind-Project/Lind-GlibC/pull/5
https://github.com/Lind-Project/nacl_repy/pull/14
https://github.com/Lind-Project/lind_project/pull/19


src/trusted/gio/gio_nacl_desc_test.c
```diff
@@ -205,7 +207,8 @@ int main(int ac,
                            NACL_ABI_O_WRONLY|
                            NACL_ABI_O_TRUNC|
                            NACL_ABI_O_CREAT,
-                           0666);
+                           0666,
+			   CAGE_TEST_ID);
```
lukpueh: Looks like there is mix in tabs and spaces here and below.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -663,6 +663,9 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
       retval = -NACL_ABI_ENOMEM;
       goto cleanup;
     }
+    /* Assign CageID to HostDir */
```
lukpueh: Would it make sense to use more mnemonic variable names so that comments can be used to explain why something happens, instead of stating the (then) obvious?

lukpueh (commented):
I fear I lack familiarity with the matter to leave anything else than superficial comments here (see inline). But the changes look straight-forward, i.e. adding cage_id (btw. what is that used for?) to function signatures and adopting the corresponding function calls. I also looked up the `Py_BuildValue` format specfication, and your changes seem right.  OTOH, I did not check if the code compiles, nor can I judge whether the changes are complete. Is there anything in particular you would like me to look at?

src/trusted/gio/gio_nacl_desc_test.c
rennergade: Thanks!

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: I updated the comment to make what I added more clear. Updating the variable names would be nice, but I'm trying to keep NaCl's original code as consistent as possible.

PR #14, opened 2019-11-23
Dup

rennergade:
This is the native client portion of Dup.

This change reflects how NaCl descriptors now work, where a new descriptor is actually made on each dup call.

This also changed the function signatures for RPC calls to reflect our caging system.


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -475,14 +475,37 @@ int32_t NaClSysDup(struct NaClAppThread *natp, int oldfd) {
...
+  new_hd->cageid = nap->cage_id;
+
+  /* Set new nacl desc as available */
+  int new_hostfd = NaClSetAvail(nap, ((struct NaClDesc *) NaClDescIoDescMake(new_hd)));
+
+  /* We've got to put that old NaClDescriptor back in there... */
+  NaClSetDesc(nap, old_hostfd, old_nd);
+
+  fd_cage_table[nap->cage_id][nap->fd] = new_hostfd;
+
```
jesings, line 508: Tell me if my summary is correct.
You're taking a NaClHostDesc, deepcopying it, then making sure that both the old and new NaClHostDesc point to the same file, then you put that new NaClHostDesc's fd into the cage table.
I'm not sure what this line does though, could you explain it?
`NaClSetDesc(nap, old_hostfd, old_nd);`

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -1749,3 +1749,14 @@ void InitializeCage(struct NaClApp *nap, int cage_id) {
...
+{
+  printf("--------------------------");
+  printf("printing desc table for nap w cageid: %d\n", nap->cage_id);
+  printf("ptr array at length %d\n", nap->desc_tbl.num_entries);
+  for (int i = 0; i < nap->desc_tbl.num_entries; i++)
+  {
+    printf("Pointer for %d 0x%" PRIXPTR "\n", i, (uintptr_t)nap->desc_tbl.ptr_array[i]);
+  }
+  printf("---------------------------");
+}
```
jesings, line 1762: This is as of yet unused?
Will it be integrated in as a verbosity option?

jesings (commented):
Looks good, your changes on dup2 are by and large well commented and straightforward, and do seem to fix the general problems with the dup2 syscall.

PR #16, opened 2020-01-23
fix race condition in fork

rennergade:


rennergade:
This is a minor fix so the child doesn't launch before cages are set up in SafePOSIX.

PR #17, opened 2020-01-31
Tlsfix

rennergade:
Minor code changes to fix a big problem.

Essentially, we want to transmit the parents TLS values to the child when spawning a thread for consistency. This effects malloc but also assumedly anything using thread local storage.


jesings:
2 lines, and more than 2 headaches in the making, I'll merge this. All this change does is clone the thread local storage of the parent for the child cages, looks great!

jesings (commented):
This code change is minimal and well documented. Sounds great!

PR #18, opened 2020-02-06
Fix exec replication, Fix Cagetable and Remove cache table

rennergade:
This PR does a few things.

- It fixes replication of FD table in SafePOSIX needed by exec
- adds some extra logging in Exec
- Fixes a cagetable error for maximum FDs
- Rips out the former cache table which is unused and was causing bugs
- Fixes some defines in NaCl globals


src/trusted/service_runtime/nacl_globals.h
```diff
@@ -42,26 +42,12 @@ EXTERN_C_BEGIN
...
-  CHILD_NUM_MAX = 256,
-  CAGING_FD_NUM  = 256,
-  ARG_LIMIT = 4096,
-  PIPE_BUF_MAX = 4906
-};
+#define  CAGING_LIB_PATH_MAX  64
+#define  FILE_DESC_MAX        1024
+#define  CAGE_MAX             1024
+#define  CHILD_NUM_MAX        256
+#define  ARG_LIMIT            4096
```
jesings, line 49: For clarity, these should ideally be written in hex or binary. I don't necessarily see the need to change it from an enum to defines, but this is just coding style and it can be done either way.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4288,14 +4279,29 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const
...
   NaClVmmapChangeProt(&nap_child->mem_map, tramp_pnum, tramp_npages, PROT_RX);
+  NaClLog(2, "Setting parent permissions to executable\n");
   NaClVmmapChangeProt(&nap->mem_map, tramp_pnum, tramp_npages, PROT_RX);
+
+  NaClLog(2, "Child mprotect\n");
   if (NaClMprotect((void *)child_start_addr, tramp_size, PROT_RX) == -1) {
+    NaClLog(LOG_FATAL, "%s\n", "child vmmap page NaClMprotect failed!");
   }
+  NaClLog(2, "Parent mprotect\n");
+
```
jesings, line 4292: Do we still want to keep these logs or were they just debugs?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4288,14 +4279,29 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const
...
   }
 
+  /* Copy fd table in SafePOSIX */
+  NaClXMutexLock(&nap->mu); 
+  NaClXMutexLock(&nap_child->mu); 
+  NaClLog(2, "Copying fd table in SafePOSIX\n");
+  lind_fork(nap_child->cage_id, nap->cage_id);
+  NaClXMutexUnlock(&nap_child->mu);
+  NaClXMutexUnlock(&nap->mu);
+
```
jesings, line 4304: This is the meat of the issue, right?

jesings (commented):
Overall, this seems simple enough and other than some stylistic issues looks good.

src/trusted/service_runtime/nacl_globals.h
rennergade: I think these in decimal makes the most sense. Stylistically this made more sense to me as well.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: Yeah, for now I think this is helpful to keep in here.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: This is the meat of the exec fds in safeposix. The meat of the real issue was the weird cache_table that I removed.

PR #19, opened 2020-02-06
Exec args fix

rennergade:
This fixes a minor error in how exec counts args and env variables.

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -4096,7 +4096,7 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const 
   
     /* Count amount of env from acquired NaCl pointer */
     uint32_t *envcountptr = sys_envp_ptr;
-    while (envcountptr[new_argc] != NULL) {
+    while (envcountptr[new_envc] != NULL) {
       new_envc++;
     }
 
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -4126,9 +4126,12 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const 
     NaClEnvCleanserDtor(&env_cleanser);
     goto fail;
   }
-  for (char **pp = sys_argv_ptr; *pp; ++pp) {
+
+  uint32_t *countptr = sys_argv_ptr;
+  while (countptr[new_argc] != NULL) {
     new_argc++;
   }
+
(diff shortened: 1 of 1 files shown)


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4093,10 +4093,13 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const
   /* set up environment, only do this if we initially were passed an environment*/
   NaClEnvCleanserCtor(&env_cleanser, 0);
   if (envp) {
-    /* count number of environment strings */
-    for (char **pp = sys_envp_ptr; *pp; ++pp) {
+  
+    /* Count amount of env from acquired NaCl pointer */
+    uint32_t *envcountptr = sys_envp_ptr;
+    while (envcountptr[new_envc] != NULL) {
```
jesings, line 4099: The previous version checked for the number of non-null char*s (64 bits) in the array. Your current version checks the number of non-null u_int32. Without being to familiar with the surrounding code, this raised some red flags. I just want to make sure that this does what you want it to do.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4126,9 +4129,13 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const
     NaClEnvCleanserDtor(&env_cleanser);
     goto fail;
   }
-  for (char **pp = sys_argv_ptr; *pp; ++pp) {
+  
+  /* Count amount of args from acquired NaCl pointer */
+  uint32_t *argcountptr = sys_argv_ptr;
+  while (argcountptr[new_argc] != NULL) {
```
jesings, line 4135: Same 32 vs 64 bit concern as above.

jesings (commented):
If the 32 vs 64 bit thing was what was wrong with the program then it all looks great. If not, you might be doing something unintended that works but changes some behavior.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: Right. That's what was giving the error!

JustinCappos:
Can we add a note indicating your assumption to the code here?  This may
save someone else (possibly even you) debugging headache in the future...

rennergade:
I have this comment a bit further up which I think clarifies why we're dealing with 32 bit pointers, but I can add a bit more.

```
 /* 
    Convert to a Sys Pointer for argv** 
    We need to turn these to uint32_t pointers for now because these are still NaCl pointers
    We'll convert to a char** later.
  */
```

PR #22, opened 2020-02-14
Develop to Master Merge

rennergade:
Native Client portion of Develop->Master Merge

Several major changes have been done during this time period including: pipe infrastructure, dup/dup2 changes, some fork modifications, and some general fixes.

Let's make sure our comments are especially clear here.


src/shared/platform/lind_platform.c
```diff
@@ -765,18 +765,25 @@ int lind_flock (int fd, int operation)
...
 }
 
 int lind_pipe2(int* pipefds, int flags){
     LIND_API_PART1;
-    callArgs = Py_BuildValue("(i[i])", LIND_safe_sys_pipe2, flags);
+    callArgs = Py_BuildValue("(i[i])", LIND_safe_fs_pipe2, flags);
     LIND_API_PART2;
     COPY_DATA(pipefds, 2*sizeof(int))
     LIND_API_PART3;
 }
```
jesings: We should clarify that pipe2 is not implemented, and anyway it should have cageid here too.

src/trusted/service_runtime/nacl_syscall_handlers_gen.py
```diff
@@ -246,7 +246,7 @@
     ('NACL_sys_lind_syscall', 'NaClSysLindSyscall', ['uint32_t callNum', 'uint32_t inNum', 'void *inArgs', 'uint32_t outNum', 'void *outArgs']),
     ('NACL_sys_fork', 'NaClSysFork', []),
     ('NACL_sys_execv', 'NaClSysExecv', ['void *path', 'void *argv']),
-    ('NACL_sys_execve', 'NaClSysExecve', ['void *path', 'void *argv', 'void *envp']),
+    ('NACL_sys_execve', 'NaClSysExecve', ['const char *path', 'void *argv', 'void *envp']),
```
jesings: You changed the type of path in execve but not execv, and you changed the type of argv in the signature in nacl_syscall_common, but not in here.

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -1750,3 +1745,14 @@ void InitializeCage(struct NaClApp *nap, int cage_id) {
...
+void print_desctbl(struct NaClApp *nap)
+{
+  printf("--------------------------");
+  printf("printing desc table for nap w cageid: %d\n", nap->cage_id);
+  printf("ptr array at length %d\n", nap->desc_tbl.num_entries);
+  for (int i = 0; i < nap->desc_tbl.num_entries; i++)
+  {
+    printf("Pointer for %d 0x%" PRIXPTR "\n", i, (uintptr_t)nap->desc_tbl.ptr_array[i]);
+  }
+  printf("---------------------------");
```
jesings: I believe the first and last printfs should either have a newline at the end or be changed to puts.

jesings (commented):
Although it might be better to comment some of the stuff that we made small changes on in here, in general, if we added a significant amount of code we commented it, and the stuff we made small changes to is left unchanged. I'd say this is generally okay to merge.

src/trusted/service_runtime/sel_ldr.c
rennergade: You're right, this is just a function I created for debug anyways. Should probably be removed for this!

rennergade (commented):
A lot of this is adding cageid all over the place, and ripping up the former pipe infrastructure. Of the major additions here I think the commenting is good.

Jonathan's suggestions should be changed but I'm fine with this.

src/trusted/service_runtime/nacl_syscall_handlers_gen.py
rennergade: This ones a little confusing because it doesn't let me do `const char*` or `char* const*` in the framework for the handler gen, so they have to be void. I'll be consistent with voids here in this file but use the actual signatures in syscall common.

jesings: Then, sorry to be such a pest, that should be commented.

rennergade:
I think this branch is good to go after the fixes I pushed.

PR #23, opened 2020-03-17
Thread PR

rennergade:
This PR consolidates our threading pipeline.

I've created an option that lives inside the nap that indicates which thread-type (MAIN, FORK, or EXEC) were launching and are able to set that accordingly.

This moves ThreadCreate, ThreadSpawn, and ThreadLauncher all into one consolidated function instead of one each for fork and main.

Most of the important changes here are in nacl_app_thread.c and sel_ldr_std.c as well as a few in nacl_syscall_common.c. A bunch of the changed files are just updating the native client test files to include the new thread syntax.

I probably, like always, need better comments here. Please let me know anything that's unclear.

### DIFF
--- a/src/trusted/service_runtime/sel_ldr.h
+++ b/src/trusted/service_runtime/sel_ldr.h
@@ -114,6 +114,12 @@ struct NaClSpringboardInfo {
   uint32_t end_addr;
 };
 
+enum NaClThreadLaunchType {
+  MAIN,
+  FORK,
+  EXEC
+};
+
 struct NaClApp {
   /*
    * children table lock children_mu is higher in the locking order than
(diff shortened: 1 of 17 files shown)


src/trusted/service_runtime/sel_ldr.h
```diff
@@ -114,6 +114,12 @@ struct NaClSpringboardInfo {
   uint32_t end_addr;
 };
 
+enum NaClThreadLaunchType {
+  MAIN,
+  FORK,
+  EXEC
+};
```
jesings: I think to make sure we don't run into naming conflicts later down the line that are hard to diagnose something like 
THREAD_LAUNCH_MAIN
THREAD_LAUNCH_FORK
THREAD_LAUNCH_EXEC
would be better.

jesings (commented):
Other than the one comment I made, this all looks fine to me, although I do wonder whether this is only because I am somewhat familiar with these functions already after having looked into how it used to work before.

src/trusted/service_runtime/sel_ldr.h
rennergade: This was a good idea, I just updated to reflect this.

PR #24, opened 2020-03-23
Fd and dup updates

rennergade:
This PR:

1. Fixes the cagetable, and fixes std stream cage initialization
2. Fixes fd allocation to use a next available system
3. Adds RPC for exec
4. Fixes how dup2 checks for existing fds

### DIFF
--- a/src/shared/platform/lind_platform.h
+++ b/src/shared/platform/lind_platform.h
@@ -82,9 +82,12 @@
 #define LIND_safe_fs_flock              54
 #define LIND_safe_fs_rename             55
 
-#define LIND_safe_fs_pipe              66
-#define LIND_safe_fs_pipe2             67
+#define LIND_safe_fs_pipe               66
+#define LIND_safe_fs_pipe2              67
 #define LIND_safe_fs_fork               68
+#define LIND_safe_fs_exec               69
+
+
 
 #define LIND_comp_cia                   105
 #define LIND_comp_call                  106
--- a/src/shared/platform/lind_platform.h
+++ b/src/shared/platform/lind_platform.h
@@ -61,7 +61,7 @@
 #define LIND_safe_net_bind              33
 #define LIND_safe_net_send              34
 #define LIND_safe_net_sendto            35
-#define LIND_safe_net_recv              362
+#define LIND_safe_net_recv              36
 #define LIND_safe_net_recvfrom          37
 #define LIND_safe_net_connect           38
 #define LIND_safe_net_listen            39
--- a/src/shared/platform/lind_platform.h
+++ b/src/shared/platform/lind_platform.h
@@ -159,5 +159,7 @@ int lind_flock (int fd, int operation);
 int lind_pipe(int* pipefds, int cageid);
 int lind_pipe2(int* pipefds, int flags, int cageid);  /* unimplemented */
 int lind_fork(int newcageid, int cageid);
+int lind_exec(int newcageid, int cageid);
+
 
 #endif /* LIND_PLATFORM_H_ */
--- a/src/trusted/service_runtime/nacl_app_thread.c
+++ b/src/trusted/service_runtime/nacl_app_thread.c
@@ -146,12 +146,13 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap) {
   
   /* duplicate file descriptor table starting at child_fd = 3 (0-2 setup previously)*/
   NaClXMutexLock(&nap_parent->mu);
-  for (int parent_fd = 0; parent_fd <= FILE_DESC_MAX; parent_fd++) {
+
+  for (int fd = 0; fd <= FILE_DESC_MAX; fd++) {
 
     /* Retrive the host fd we had stored in the Cage Table for the parent */
-    int parent_host_fd = fd_cage_table[nap_parent->cage_id][parent_fd];
+    int parent_host_fd = fd_cage_table[nap_parent->cage_id][fd];
     if (parent_host_fd == -1) {
-      fd_cage_table[nap_child->cage_id][nap_child->fd++] = -1;
+      fd_cage_table[nap_child->cage_id][fd] = -1;
       continue;
     }
     /* Retrieve Parent NaCl Descriptor based on current child fd in the parent */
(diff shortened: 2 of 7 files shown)


src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -117,7 +117,7 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap) {
     NaClXMutexUnlock(&nap_parent->children_mu);
   }
 
-  NaClAppInitialDescriptorHookup(nap_child);
+  //NaClAppInitialDescriptorHookup(nap_child);
```
JustinCappos: Is this intended?  If so, do you want to describe why you commented this but left the code?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -146,12 +146,13 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap) {
...
-  for (int parent_fd = nap_child->fd; parent_fd <= nap_parent->fd; parent_fd++) {
+
+  for (int fd = 0; fd <= FILE_DESC_MAX; fd++) {
 
     /* Retrive the host fd we had stored in the Cage Table for the parent */
-    int parent_host_fd = fd_cage_table[nap_parent->cage_id][parent_fd];
-    if (parent_host_fd == 0) {
-      fd_cage_table[nap_child->cage_id][nap_child->fd++] = 0;
+    int parent_host_fd = fd_cage_table[nap_parent->cage_id][fd];
+    if (parent_host_fd == -1) {
```
JustinCappos: I presume this is being set so that you don't confuse it with stdin.  Would a special #defined macro value be better?
```suggestion
    // Using -1 (an invalid fd) to avoid confusing this with a valid fd.
    if (parent_host_fd == -1) {
```

JustinCappos (commented):
I'm leaving a few comments with some minor stylistic things that it may be worth considering.  @rennergade You can take or leave these, but if you do this sort of work throughout, it will be slightly slower to code, but in my experience it is much, much faster to debug / read later.

src/trusted/service_runtime/nacl_app_thread.c
rennergade: Very much appreciate the review/comments! This was just a mistake, I was testing moving the initial descriptor init from being a separate function to being part of the main fd transfer that's below. It worked, and then I forgot to delete.

src/trusted/service_runtime/nacl_app_thread.c
rennergade: A macro definitely seems like a good idea!

src/shared/platform/lind_platform.h
```diff
@@ -82,9 +82,12 @@
 #define LIND_safe_fs_flock              54
 #define LIND_safe_fs_rename             55
 
-#define LIND_safe_fs_pipe              66
-#define LIND_safe_fs_pipe2             67
+#define LIND_safe_fs_pipe               66
+#define LIND_safe_fs_pipe2              67
```
jesings, line 86: I'm pretty sure I fixed this in my getpid branch, we will have a merge conflict then, I'll deal with it though.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -146,12 +145,13 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap) {
...
 
     /* Retrive the host fd we had stored in the Cage Table for the parent */
-    int parent_host_fd = fd_cage_table[nap_parent->cage_id][parent_fd];
-    if (parent_host_fd == 0) {
-      fd_cage_table[nap_child->cage_id][nap_child->fd++] = 0;
+    int parent_host_fd = fd_cage_table[nap_parent->cage_id][fd];
+    if (parent_host_fd == NACL_BAD_FD) {
+      fd_cage_table[nap_child->cage_id][fd] = NACL_BAD_FD;
       continue;
     }
```
jesings, line 156: Don't you want a special case for fds 0 - 2? I don't completely understand the changes in general (that is, the changes in the selected region), if you could explain them that would help me understand this pr better.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -465,19 +465,13 @@ int32_t NaClSysDup(struct NaClAppThread *natp, int oldfd) {
...
 
-  /* check for closed fds */
-  if (oldfd < 0) {
-    ret = -NACL_ABI_EBADF;
-    goto out;
-  }
+  old_hostfd = fd_cage_table[nap->cage_id][oldfd];
 
-  if (nap->fd > FILE_DESC_MAX) {
+  if (old_hostfd < 0) {
```
jesings, line 470: Do you no longer care about the max number for fds?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -529,23 +523,18 @@ int32_t NaClSysDup2(struct NaClAppThread  *natp,
...
-    goto out;
-  }
 
   /* Get old host fd from cage table, and use that to get nacl descriptor */
   old_hostfd = fd_cage_table[nap->cage_id][oldfd];
+
+  if (old_hostfd < 0) {
+    ret = -NACL_ABI_EBADF;
+    goto out;
+  }
```
jesings, line 537: Do you no longer care about max fd here either?

jesings (commented):
I think your fd and dup fixes were pretty straightforward aside from the comments I made, and your exec rpc seems fine, some of the architecture of this stuff before you fixed it leaves me baffled though, but I guess lind is totally built on a house of cards, so...

src/trusted/service_runtime/nacl_app_thread.c
rennergade: This has to do with the commented out function Justin found above. I removed the separate initial descriptor setup function and replaced it with us copying over all descriptors. Because of dup functions we want new cages to inherit the previous cage's descriptors and not set up a brand new 0-2.

rennergade: Other than that, I got rid of the confusing parent and child counters and replaced it with just one fd counter that goes over all possible fds. I know this is O(n) and isn't great, but with the better fd allocation system this is the easiest way right now to make sure we copy all descriptors.

I should probably make an issue about both the allocation and copying that there are some O(n) functions that could be made more efficient when we're looking to increase speed. OTOH having an n=1024 of something that's not called very often isn't horrible.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: The new fd ticketing system makes this not necessary anymore. Though your comment did make me realize the ticketer should return the ABI code for bad fd and not my bad fd marker.

PR #26, opened 2020-03-28
Wait debug

rennergade:
I somehow fast tracked this Pull Request accidentally, so doing it again for a code review.

Here I fixed a problem with wait where a children_id table was being used in a linear fashion, which caused a bug due to children not exiting in any pre-known order. The fix here gets rid of that table and just uses the dynamic array of ptrs to child naps. It also gets rid of some other unused nap variables.

### DIFF
--- a/src/trusted/service_runtime/nacl_app_thread.c
+++ b/src/trusted/service_runtime/nacl_app_thread.c
@@ -103,12 +103,7 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap) {
     }
     NaClLog(1, "[nap %d] incrementing num_children\n", nap_arr[i]->cage_id);
     
-    printf("%d number of children \n", nap_arr[i]->num_children);
-
-    nap_arr[i]->children_ids[nap_arr[i]->num_children++] = nap_child->cage_id;
-
-
-    printf("%d number of children \n", nap_arr[i]->num_children);
+    nap_arr[i]->num_children++];
 
     if (nap_arr[i]->num_children > CHILD_NUM_MAX) {
       NaClLog(LOG_FATAL, "[nap %u] child_idx > %d\n", nap_arr[i]->cage_id, CHILD_NUM_MAX);
@@ -330,18 +325,6 @@ void WINAPI NaClAppThreadLauncher(void *state) {
   }
 }
 
-static INLINE int GetChildIdx(const volatile sig_atomic_t *id_list, int nmemb, int cage_id) {
-  int ret;
-  /* return the index if match found */
-  for (ret = 0; ret < nmemb; ret++) {
-    if (cage_id == id_list[ret]) {
-      return ret;
-    }
-  }
-  /* otherwise return an error value */
-  return -1;
-}
-
 /*
  * preconditions:
  * * natp must be thread_self(), called while holding no locks.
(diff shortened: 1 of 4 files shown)


src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -323,18 +325,6 @@ void WINAPI NaClAppThreadLauncher(void *state) {
   }
 }
 
-static INLINE int GetChildIdx(const volatile sig_atomic_t *id_list, int nmemb, int cage_id) {
```
jesings, line 326: I just wanted to comment on concurrency concerns.
The child ids are in an array of sig_atomic_t. Presumably this access is atomic for a reason. Are you sure you have eliminated any possibility for concurrency issues by eliminating this atomic component?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4406,17 +4406,17 @@ int32_t NaClSysWaitpid(struct NaClAppThread *natp,
...
     goto out;
   }
 
   /*
-   * explicit child pid given
+   * WAITPID: explicit child pid given
    */
-  if (cage_id > 0 && cage_id <= pid_max) {
+  if (pid > 0 && pid <= pid_max) {
+    int cage_id = pid;
```
jesings, line 4419: What is the purpose of creating a new variable, cage_id here? Can't you just change the below calls to use pid as an arg?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4433,51 +4433,57 @@ int32_t NaClSysWaitpid(struct NaClAppThread *natp,
...
+    ret = pid;
     goto out;
   }
 
   /*
-   * TODO: implement pid == WAIT_ANY_PG (0) and pid == WAIT_ANY (-1) behavior
-   *
-   * -jp
+   * WAIT or WAITPID(-1)
+   * TODO: implement pid == WAIT_ANY_PG (0), we currently don't deal with process groups
```
jesings, line 4442: We still have WAIT_ANY (-1) unimplemented, right?

jesings: If so make a note of that

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4433,51 +4433,57 @@ int32_t NaClSysWaitpid(struct NaClAppThread *natp,
...
+          }
+        }
+    
+        NaClXCondVarBroadcast(&nap->children_cv);
+
+        NaClXMutexUnlock(&nap->children_mu);
       }
-      if (++cur_idx > CHILD_NUM_MAX) {
-        cur_idx = 0;
-      }
```
jesings, line 4470: I notice that cur_idx is volatile. Presumably that means it can be changed from elsewhere, and presumably that's what the mutex code is protecting against. I do want to make sure that in creating your new look you have removed the corresponding (external?) code that would change cur_idx, if this is not the only place it can be modified. Also, I want to make sure that if you have done this it still makes sense to have the same amount of mutex code.

jesings (commented):
There do seem to be some kinds of concurrency that you have removed, but maybe you removed the need for them to. I just want to be sure of this before we pull.

src/trusted/service_runtime/nacl_app_thread.c
rennergade: Yeah, exactly. At this point we don't use this id_list structure at all, since it was redundant. I understand why it was added somewhat, as a way to visualize the waiting process, but I think my way is much cleaner as well as not having to deal with more concurrency.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: cur_idx was used to index against the child_ids list that I removed. I'm not exactly sure why it was volatile, since the index itself couldn't be changed from elsewhere, but the ids could, so that may have been the confusion. Either way its not necessary anymore.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: So I did it this way to try to keep some consistency with the WAIT_ANY protocol that also uses cage_id. In my head `pid` is a term we're using in the POSIX context where `cage_id` is what we're using to deal with our internal structures. Maybe its not useful, or I should note this.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: The second part of waitpid() here is what deals with WAIT_ANY, which is the same as a normal wait().

rennergade:
@jesings has given me the green light to merge this.

PR #29, opened 2020-04-15
Timeinfo

rennergade:
Updates lind to use a flag "-t" to get timeinfo without extraneous logging.


jesings (commented):
This is a pretty trivial PR, I have no comments. Looks good!

PR #30, opened 2020-04-29
Mmap

jesings:


src/shared/platform/posix/nacl_host_desc.c
```diff
@@ -183,7 +186,14 @@ uintptr_t NaClHostDescMap(struct NaClHostDesc *d,
    */
   tmp_prot = host_prot & ~PROT_EXEC;
   need_exec = (0 != (PROT_EXEC & host_prot));
-  map_addr = mmap(start_addr, len, tmp_prot, host_flags, desc, offset);
+  if(!start_addr){
+    NaClLog(LOG_FATAL,
+            "NaClHostDescMap: start_addr cannot be NULL.\n");
+  }
+  whichcage = d ? d->cageid : 0;
```
rennergade, line 196: I think a bit of commenting here would be helpful for the future, I understand what's going on but we had talked about it.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1527,12 +1527,13 @@ static int32_t MunmapInternal(struct NaClApp *nap, uintptr_t sysaddr, size_t len
    * zero-filled pages, which should be copy-on-write and thus
    * relatively cheap.  Do not open up an address space hole.
    */
-  if (MAP_FAILED == mmap((void *) sysaddr,
-                         length,
-                         PROT_NONE,
-                         MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED,
-                         -1,
-                         (off_t) 0)) {
+  if (MAP_FAILED == lind_mmap((void *) sysaddr,
```
rennergade, line 1530: So we're still mmaping zeros here again to munmap, but you added munmap in SafePOSIX. I'm a bit confused as to what the goal is still.

rennergade (commented):
Mostly standard, had a few questions.

src/shared/platform/posix/nacl_host_desc.c
jesings: Okay will do.

src/trusted/service_runtime/nacl_syscall_common.c
jesings: We are still mmapping 0s, I added munmap to SafePOSIX to complete the API, because I think we'd really have to futz around with NaCl's internal memory map in order to change this behavior and not have lots of seg faults or a huge memory leak.
We can do that if you think it's necessary though. Although if we do that that can be in a separate branch/merge I think.

src/shared/platform/posix/nacl_host_desc.c
jesings: Added!

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: Yeah thats fine. We should continue to do what they implemented, I just want to make sure we don't get confused about this in the future.

PR #31, opened 2020-05-20
Mmapfix

jesings:


src/shared/platform/posix/nacl_host_desc.c
```diff
@@ -203,7 +203,7 @@ uintptr_t NaClHostDescMap(struct NaClHostDesc *d,
    */
   mapbottom = lind_mmap(start_addr, len, tmp_prot, host_flags, desc, offset, whichcage);
   //MAP_FAILED is -1, so bitwise or will have no effect
-  map_addr = (void*) (topbits | (long) mapbottom);
+  map_addr = (void*) (mapbottom == (unsigned int) -1 ? (unsigned long) -1L : topbits | (unsigned long) mapbottom);
```
rennergade, line 209: Also comment here. In general ternary's can be confusing so it should be explained, especially since it was causing us an error here.

PR #34, opened 2020-08-02
Added Exit() Interface & RPC

kliu513:


rennergade:
This looks good and is ready to merge, but needs to be merged with both repy and glibc. Was there supposed to be a new GLIBC PR as well?

PR #35, opened 2020-09-24
Getpid

jesings:
I see that the 31 is not aligned in the syscall list, but if there are any other fixes you see I'll include them in the same commit


rennergade:
There seems to be some mmap stuff in here, and a bunch of weird edits to the lind_platform file, including getting rid of PythonFunc0, which I think we need?

jesings:
I don't see what you mean by mmap stuff.
The edits to the lind platform files are mostly a few syntactic choices (i.e. opening curly brace on the same line) that were different from the rest of the file.
I got rid of PythonFunc1, not PythonFunc0, which I believe was a variant of PythonFunc that I believe was for sending objects for packed structs to be returned but is deprecated and unused. If you think that belongs in a different commit or not at all I can take that out, I'm not 100% sure why I did that 4 months ago in this commit. I think I removed it here because some deprecated getpid path used it.

rennergade:
If you look in the commits there are some mmap labeled commits. 

Yeah, let's leave that function in for now. It seems like we're going to have to rework some things on the rpc side so I'd rather do it in one sweep to avoid confusion.

jesings:
Oh the mmap stuff is because I branched getpid off of mmap but those mmap changes should already have been merged into develop separately so although they're listed here nothing will actually be merged. I'll leave the CallPythonFunc1 in for now, I'll let you know when I've made the commit reverting that and fixing the alignment in the header.

PR #38, opened 2020-11-11
Only copy stack from stack pointer, drastically reducing copy time in…

jesings:
… fork


rennergade (approved):
This all looks great.

PR #39, opened 2020-11-11
Remove extra syscalls from NaCl

rennergade:
Removed fstat in open and mmap
Changed open to create NaClDesc for directories instead of NaClDir, and getdents to go directly to a safeposix call to avoid using NaCl dirs. This allows the removal of stat in open.


jesings (commented):
Looks good, if a bit complicated as you're getting rid of a lot of stuff. I am wondering how you tell whether a file is a directory or not within open without doing an fxstat, as I don't see you determining that elsewhere. Did this difference just not matter?

rennergade:
> Looks good, if a bit complicated as you're getting rid of a lot of stuff. I am wondering how you tell whether a file is a directory or not within open without doing an fxstat, as I don't see you determining that elsewhere. Did this difference just not matter?

We're basically removing dir's on the NaCl side completely, so we don't need this fstat anymore. Errors are handled internally by safeposix so it was redundant.

PR #40, opened 2020-11-13
Fix NaClFdToRepyFD in lind_syscalls to fix sort/file problems & fix open to return -1 when failed.

rennergade:


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -654,7 +654,7 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
                     int                   flags,
                     int                   mode) {
   struct NaClApp       *nap = natp->nap;
-  uint32_t             retval = -NACL_ABI_EINVAL;
+  int32_t              retval = -NACL_ABI_EINVAL;
```
jesings: An int rather than an int32_t would probably make more sense to me here.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -722,10 +722,14 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
...
           nap->cage_id, (uintptr_t) hd, path, flags, mode, retval);
-  if (!retval) {
+  if (retval >= 0) {
     retval = NaClSetAvail(nap, ((struct NaClDesc *) NaClDescIoDescMake(hd)));
     NaClLog(1, "Entered into open file table at %d\n", retval);
   }
+  else {
+    NaClLog(1, "Open returned error %d\n", retval);
+    return -NACL_ABI_EPERM;
+  }
```
jesings: I'd put this above where you currently have your if block and remove the if statement, so something like
```
if(retval < 0) {
    NaClLog(1, "Open returned error %d\n", retval);
    return -NACL_ABI_EPERM;
}
retval = NaClSetAvail(nap, ((struct NaClDesc *) NaClDescIoDescMake(hd)));
NaClLog(1, "Entered into open file table at %d\n", retval);
```
Not a big deal but seems a little cleaner to me.

jesings (commented):
Looks pretty much good to me.

rennergade:
Ok, great comments, I made those changes.

PR #42, opened 2021-01-08
Bugfixes and removal of vestigial globals

jesings:
Many small things were done:
Fixed parent-exit bug by adding a global mutex/counter/condvar.
Removed Joey's previous attempt at solving this problem by making all cages children of the initial cage!?
Removed globals that had been set up to track initial cage.
Fixed mmap child duplicate entry bug.
Fixed env cleanser bug.
Removed NaClPatchAddr.
Fixed bug in LindPythonFinalize wherein the lind.mix function finalize was being called twice, causing SIGTERM to be raised from within CPython.
Fixed vmmap child desc bug.
Removed unused variables.


src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -79,46 +72,40 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap) {
...
+  cagecount++;
+  NaClXMutexUnlock(&ccmut);
+
+  NaClXMutexLock(&nap_parent->children_mu);
   /*
    * increment fork generation count and generate child
    * cage_id (both master and parent mutexes need to be held
    */
-  InitializeCage(nap_child, nap_master->cage_id + ++fork_num);
+  InitializeCage(nap_child, 1 +  ++fork_num);
```
rennergade: Should we be fixing this to start from cage 0 at this point? I think the "mystery" Cage 0 was fixed.

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -1731,7 +1731,7 @@ void NaClCopyExecutionContext(struct NaClApp *nap_parent, struct NaClApp *nap_ch
   NaClLog(2, "Initializing arch switcher\n");
   NaClInitSwitchToApp(nap_child);
   NaClLog(2, "Installing trampoline\n");
-  NaClLoadTrampoline(nap_child);
+  //NaClLoadTrampoline(nap_child); handled in NaClLoadFileAslr
```
rennergade, line 1738: Should this still be commented?

src/trusted/service_runtime/nacl_globals.h
```diff
@@ -107,6 +105,7 @@ void  NaClGlobalModuleFini(void);
 void NaClInitGlobals(void);
 
 static INLINE void NaClPatchAddr(uintptr_t child_bits, uintptr_t parent_bits, uintptr_t *start, size_t size) {
+  return;
```
rennergade: I think we should remove this entirely before we merge.

rennergade (commented):
I think we should probably test removing PatchAddr completely before we merge this. Doesn't make sense to merge a half-measure.

Also may make sense to add a fix to VMmap copy bug to this PR.

Otherwise looks awesome. There's a ton of great work in here.

src/trusted/service_runtime/sel_ldr.c
jesings: I think so, as I'm not entirely sure whether we'll want to stop going through most of NaClLoadFileAslr in forks or not, as it does certain things twice that we can't remove so easily here.

src/trusted/service_runtime/nacl_app_thread.c
jesings: I mean in posix, the init process has a pid 1, so I think 1 being the first pid is fine.

rennergade: Hmm, maybe we should make 1 a define here for clarity then.

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -1619,27 +1616,34 @@ void NaClCopyDynamicTextAndVmmap(struct NaClApp *nap_parent, struct NaClApp *nap
...
+            NaClLog(LOG_FATAL, "%s\n", "parent vmmap page NaClMprotect failed!");
+          }
         }
+      } else {
+        NaClVmmapAddWithOverwrite(&nap_child->mem_map,
+                                  entry->page_num,
+                                  entry->npages,
+                                  entry->prot,
+                                  (entry->flags | MAP_ANON_PRIV) & ~NACL_ABI_MAP_SHARED,
+                                  entry->desc,
```
rennergade, line 1644: Is this supposed to be the child desc?

src/trusted/service_runtime/sel_ldr.c
rennergade: Still commented?

rennergade (commented):
Really great. Will close a few issues and makes things faster/easier. Just a couple comments.

src/trusted/service_runtime/sel_ldr.c
jesings: Nope, this one is just for the text_shm desc, which is a Shm desc and is not an io desc, and therefore has no associated fd, and also no associated cage, so nothing needs to be changed about which desc is associated on fork.

src/trusted/service_runtime/sel_ldr.c
jesings: Yep. I think that it makes sense as we probably want to not use most of NaClLoadFileAslr on fork and therefore we will probably want to uncomment this line in the not too distant future.

PR #44, opened 2021-01-29
Fixed network call fd handling in lind, also multithreading fix

jesings:


rennergade (approved):
This will work well for now. Thanks!

PR #64, opened 2021-09-18
Moved socketpair syscall to the new path.

youngfalcon:


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -3137,6 +3137,55 @@ int32_t NaClSysImcSocketPair(struct NaClAppThread *natp,
...
+  struct NaClApp          *nap = natp->nap;
+  void                    *xchangedata = NULL;
+  int32_t                 retval;
+
+  NaClLog(2, "Cage %d Entered NaClSysSocketPair(0x%08"NACL_PRIxPTR", "
+           "%d, %d, %d, 0x%08"NACL_PRIx32")\n",
+           nap->cage_id, (uintptr_t)natp, domain, type, protocol, (uintptr_t)fds);
+
+  //Preprocessing start
+  xchangedata = malloc(sizeof(struct NaClHostDesc)*2);
```
rennergade: This whole function could be much neater. I know its just basically copying the pre/post processes around the RPC call, but it would be better if you organized it better. Using a more descriptive variable than "xchangedata" and things like moving the declarations to the beginning would be a good start.

rennergade (commented):
Mostly great, but I did a bit of nitpicking.

src/trusted/service_runtime/nacl_syscall_common.c
jesings: I agree

rennergade (approved):
Cool. Changes are very clean. Looks great!

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -3137,6 +3137,55 @@ int32_t NaClSysImcSocketPair(struct NaClAppThread *natp,
...
+
+    hd->cageid = nap->cage_id;
+    nacl_fd = NaClSetAvail(nap, ((struct NaClDesc *) NaClDescIoDescMake(hd)));
+    user_fd = NextFd(nap->cage_id);
+    fd_cage_table[nap->cage_id][user_fd] = nacl_fd;
+  }
+  //Postprocessing end
+
+  NaClLog(2, "NaClSysSocketPair: returning %d\n", retval);
+
```
jesings, line 3182: From what I understand you'll need to replace the 2 fds in the array with the system fds you create for them (what you assign to user_fd), so that the user program has the right fds to use

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -3137,6 +3137,55 @@ int32_t NaClSysImcSocketPair(struct NaClAppThread *natp,
...
+           "%d, %d, %d, 0x%08"NACL_PRIx32")\n",
+           nap->cage_id, (uintptr_t)natp, domain, type, protocol, (uintptr_t)fds);
+
+  //Preprocessing start
+  hd_struct = malloc(sizeof(struct NaClHostDesc)*2);
+  if (!hd_struct) {
+    retval = -NACL_ABI_ENOMEM;
+    return retval;
+  }
+  //Preprocessing end
```
jesings: The *processing end/start comments are kind of clutter

jesings (commented):
A couple more changes I think could be made, but those should be simple and overall this looks good!

src/trusted/service_runtime/nacl_syscall_common.c
youngfalcon: I think you are right, will remove them.

rennergade:
Looks good! Think this is ready.

kapkic (approved):
LGTM as well

PR #66, opened 2021-09-22
Moved Getsockname call.

kapkic:


src/shared/platform/lind_platform.c
```diff
@@ -790,3 +790,15 @@ int lind_socket (int domain, int type, int protocol, int cageid)
     LIND_API_PART2;
     LIND_API_PART3;
 }
+
+int lind_getsockname (int sockfd, socklen_t* addrlen_in, struct sockaddr * addr, socklen_t * addrlen_out, int cageid)
+{
+    LIND_API_PART1;
+    callArgs = Py_BuildValue("(i[iii])", LIND_safe_net_getsockname, sockfd, *addrlen_in, cageid);
+    LIND_API_PART2;
+    COPY_DATA(addr, *addrlen_in)
```
jesings, line 799: only _len bytes should be copied

jesings (commented):
One small change to make

src/shared/platform/lind_platform.c
kapkic: Done.
`COPY_DATA(addr, _len)`

PR #69, opened 2021-09-27
Oxidization

jesings:


rennergade (approved):
Usually I would say this is good, but really it's awesome. This RPC interface is so much nicer than the old crazy one.

Approved.

PR #72, opened 2021-10-14
Fork Fix

rennergade:
This changes around the order of how we setup fork so that the cage is ready in RustPOSIX when we launch from NaCl.


youngfalcon (approved):
Looks good to me.

jesings (approved):
After the fix this looks good

PR #73, opened 2021-10-15
Exec id fix

rennergade:
This is the same thing I just did to fork, where I move up the RPC and call exit if theres a problem making the thread.


jesings (approved):
Still looks good!

PR #74, opened 2021-11-16
Time monotonic

rennergade:
Finally merging changes that switches the `-t` mechanism to use monotonic timing.


jesings (approved):
Changes look totally fine! I think it's probably worth making sure to test the native stuff with the monotonic clock too just for fairness.

PR #76, opened 2021-12-09
Break_addr fix

rennergade:
This rearranges thread launching to give threads a launch type earlier in execution, its also a bit cleaner. This enables us later on to set the break address for a child to the parent address if it's a fork.


youngfalcon (approved):

jesings (approved):
Looks fine, you did change some files which we really haven't been using/caring about like the nacl internal tests or sel_main_chrome but that's fine.

PR #77, opened 2021-12-14
Increase IO Efficiency for Read/Write in NaCl

rennergade:
- Gets rid of/bypasses some locks that are not necessary due to RustPOSIX
- Gets rid of VMIOWillStart/Stop which is also unnecessary for us
- Rearranges some logging so it isn't slow for no reason

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -1057,7 +1057,7 @@ int32_t NaClSysWrite(struct NaClAppThread *natp,
 
   sysaddr = NaClUserToSysAddrRange(nap, (uintptr_t) buf, count);
   if (kNaClBadAddress == sysaddr) {
-    // NaClDescUnref(ndp);
+    NaClDescUnref(ndp);
     retval = -NACL_ABI_EFAULT;
     goto out;
   }
@@ -1085,7 +1085,7 @@ int32_t NaClSysWrite(struct NaClAppThread *natp,
 
   write_result = ((struct NaClDescVtbl const *)ndp->base.vtbl)->Write(ndp, (void *)sysaddr, count);
 
-  // NaClDescUnref(ndp);
+  NaClDescUnref(ndp);
 
   /* This cast is safe because we clamped count above.*/
   retval = (int32_t)write_result;
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -895,7 +895,9 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
     goto out;
   }
 
-  ndp = NaClGetDesc(nap, fd);
+  // ndp = NaClGetDesc(nap, fd);
+  ndp = (struct NaClDesc *) DynArrayGet(&nap->desc_tbl, fd);
+
   NaClLog(2, " ndp = %"NACL_PRIxPTR"\n", (uintptr_t) ndp);
   if (!ndp) {
     retval = -NACL_ABI_EBADF;
@@ -937,7 +939,7 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
   } else {
     NaClLog(4, "read returned %"NACL_PRIdS"\n", read_result);
   }
-  NaClDescUnref(ndp);
+  // NaClDescUnref(ndp);
 
   /* This cast is safe because we clamped count above.*/
   retval = (int32_t) read_result;
@@ -1048,7 +1050,9 @@ int32_t NaClSysWrite(struct NaClAppThread *natp,
     goto out;
   }
 
-  ndp = NaClGetDesc(nap, fd);
+  // ndp = NaClGetDesc(nap, fd);
+  ndp = (struct NaClDesc *) DynArrayGet(&nap->desc_tbl, fd);
+
   NaClLog(2, " ndp = %"NACL_PRIxPTR"\n", (uintptr_t) ndp);
   if (!ndp) {
     retval = -NACL_ABI_EBADF;
@@ -1085,7 +1089,7 @@ int32_t NaClSysWrite(struct NaClAppThread *natp,
 
   write_result = ((struct NaClDescVtbl const *)ndp->base.vtbl)->Write(ndp, (void *)sysaddr, count);
 
-  NaClDescUnref(ndp);
+  // NaClDescUnref(ndp);
 
   /* This cast is safe because we clamped count above.*/
   retval = (int32_t)write_result;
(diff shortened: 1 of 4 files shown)


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -897,7 +895,8 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
     goto out;
   }
 
-  ndp = NaClGetDesc(nap, fd);
+  ndp = (struct NaClDesc *) DynArrayGet(&nap->desc_tbl, fd);
+  /* Use DynArrayGet to bypass GetDesc locks for fast IO */
```
jesings: I think this creates a race condition with the dynarr length if something tries to set the dynarr length but I think we're ok with that race condition?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -920,33 +919,24 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
...
   }
 
-  NaClVmIoWillStart(nap,
-                    (uint32_t) (uintptr_t) buf,
-                    (uint32_t) (((uintptr_t) buf) + count - 1));
   read_result = ((struct NaClDescVtbl const *)ndp->base.vtbl)->Read(ndp, (void *)sysaddr, count);
 
-  NaClVmIoHasEnded(nap,
-                    (uint32_t) (uintptr_t) buf,
-                    (uint32_t) (((uintptr_t) buf) + count - 1));
```
jesings: I actually don't think we can remove these. What these do is mark which memory addresses are currently being written to/read to in order to prevent them from being mmapped/mprotected over. I'm pretty sure that removing these would allow the user to write to memory they should not be able to under some conditions.

src/trusted/service_runtime/nacl_syscall_common.c
jesings: I also think this is supposed to add a new ref to prevent the desc from being freed in between, this may or may not be something we need to worry about.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -920,33 +919,24 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
...
+    if (log_bytes > kdefault_io_buffer_bytes_to_log) {
+      log_bytes = kdefault_io_buffer_bytes_to_log;
+      ellipsis = "...";
     }
     NaClLog(8, "read result: %.*s%s\n",
             (int) log_bytes, (char *) sysaddr, ellipsis);
   } else {
     NaClLog(4, "read returned %"NACL_PRIdS"\n", read_result);
   }
-  NaClDescUnref(ndp);
```
jesings, line 949: Same as my comment on the DynArrayGet

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1059,7 +1047,8 @@ int32_t NaClSysWrite(struct NaClAppThread *natp,
     goto out;
   }
 
-  ndp = NaClGetDesc(nap, fd);
+  ndp = (struct NaClDesc *) DynArrayGet(&nap->desc_tbl, fd);
+  /* Use DynArrayGet to bypass GetDesc locks for fast IO */
```
jesings: Same concern as above

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1093,16 +1082,8 @@ int32_t NaClSysWrite(struct NaClAppThread *natp,
...
           d, (int)log_bytes, (char *)sysaddr, ellipsis, count);
 
-  NaClVmIoWillStart(nap,
-                    (uint32_t)(uintptr_t)buf,
-                    (uint32_t)(((uintptr_t)buf) + count - 1));
-  write_result = ((struct NaClDescVtbl const *)ndp->base.vtbl)->Write(ndp, (void *)sysaddr, count);
-
-  NaClVmIoHasEnded(nap,
-                   (uint32_t)(uintptr_t)buf,
-                   (uint32_t)(((uintptr_t)buf) + count - 1));
```
jesings: Same concern as above

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -625,9 +625,7 @@ struct NaClDesc *NaClGetDesc(struct NaClApp *nap,
                              int            d) {
   struct NaClDesc *res;
 
-  NaClFastMutexLock(&nap->desc_mu);
   res = NaClGetDescMu(nap, d);
-  NaClFastMutexUnlock(&nap->desc_mu);
   return res;
 }
```
jesings, line 639: Why not just replace the calls with calls to NaClGetDescMu?

jesings (commented):
I'm not so sure that we preserve some of the guarantees of NaCl by doing this even though it does speed things up significantly.

rennergade:
@jesings I believe I have updated this to either relieve your worries or we've talked about the issues ie VMIO only applying to windows.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -941,6 +946,7 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
   /* This cast is safe because we clamped count above.*/
   retval = (int32_t) read_result;
 out:
+  NaClFastMutexUnlock(&nap->desc_mu);
```
jesings, line 949: Hmm... so unref wasn't being called here at out, now I'm not as sure about any of this... that means the ref increment was actually persisting, but until where?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1057,7 +1066,6 @@ int32_t NaClSysWrite(struct NaClAppThread *natp,
 
   sysaddr = NaClUserToSysAddrRange(nap, (uintptr_t) buf, count);
   if (kNaClBadAddress == sysaddr) {
-    NaClDescUnref(ndp);
```
jesings, line 1071: Same concern about it not unrefing unless there's an error here

src/trusted/service_runtime/sel_ldr.h
```diff
@@ -854,6 +857,11 @@ void NaClVmHoleClosingMu(struct NaClApp *nap);
  * region..
  */
 
+/* Lind
+ * We're not only using Linux so it's fine to disregard these operations which are
```
jesings: remove the not, typo

jesings (commented):
The VmI/O stuff looks great. I had one lingering concern about the desc stuff.

rennergade:
I fixed the typo and we discussed your other comments, you had missed the other unref that was removed.

PR #78, opened 2021-12-28
(F/L)Stat fix

rennergade:
This PR fixes some problems that exist with stat calls.

- Fixes ordering and size of nlink in NaCl struct
- Reverts RPC calls
- Fixes NaClAbiStatHostDescStatXlateCtor, a function that was clobbering the stat results

### DIFF
--- a/src/trusted/desc/win/nacl_desc.c
+++ b/src/trusted/desc/win/nacl_desc.c
@@ -40,16 +40,9 @@ int32_t NaClAbiStatHostDescStatXlateCtor(struct nacl_abi_stat   *dst,
     return -NACL_ABI_EOVERFLOW;
   }
 
-  dst->nacl_abi_st_dev = 0;
-#if defined(NACL_MASK_INODES)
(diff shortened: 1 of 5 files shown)


src/trusted/desc/win/nacl_desc.c
```diff
@@ -40,16 +40,9 @@ int32_t NaClAbiStatHostDescStatXlateCtor(struct nacl_abi_stat   *dst,
     return -NACL_ABI_EOVERFLOW;
   }
 
-  dst->nacl_abi_st_dev = 0;
-#if defined(NACL_MASK_INODES)
```
JustinCappos, line 44: Why are we dropping this?

rennergade: This was designed in NaCl to mask host inodes on stats. Since we're now returning our own "faked" rustposix inodes, it's fine to return them to the user.

jesings (approved):
These changes look good. I'm not thrilled that we have to do this, but given that we do this seems fine.

youngfalcon (approved):

PR #80, opened 2022-01-05
Fix Argument Passing

rennergade:
This ended up being a bit simpler than I thought.

What this PR does:
1 - Switches the way Execve and Execv interact, with execve now parsing new environment variables and then passing to execv instead of vice versa.

2 - Appends "NACLENV_" to new incoming args in Execve so that they pass through the environment cleanser, and sends those args to the cleanser correctly.

This way Execv keeps the old args, but Execve replaces them with new args, even if that's NULL.

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -4137,20 +4137,29 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const 
     for (int i = 0; i < new_envc; i++) {
       char *env = (void *)NaClUserToSysAddr(nap, (uintptr_t)sys_envp_ptr[i]);
       env = (uintptr_t)env == kNaClBadAddress ? 0 : env;
-      new_envp[i] = env ? strdup(env) : 0;
+
       if (!env) {
+        new_envp[i] = 0;
         break;
       }
+      else {
+        new_envp[i] = malloc(NACL_ENV_PREFIX_LENGTH + strlen(env) + 1);
+        strcpy(new_envp[i], NACL_ENV_PREFIX);
+        strcat(new_envp[i], env);
+      } 
+    
     }
   }
 
   new_envp[new_envc] = 0;
+
   /* We've already cleaned the native environment, so just supply extra args from this syscall */
-  if (!NaClEnvCleanserInit(&env_cleanser, environ, (char const *const *)new_envp)) {
+  if (!NaClEnvCleanserInit(&env_cleanser, (char const *const *)new_envp, 0)) {
     NaClLog(LOG_ERROR, "%s\n", "Failed to initialize environment cleanser");
     NaClEnvCleanserDtor(&env_cleanser);
     goto fail;
   }
+
   nap->clean_environ = NaClEnvCleanserEnvironment(&env_cleanser);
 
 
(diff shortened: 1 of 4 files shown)


youngfalcon (approved):
Looks good to me.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4163,24 +4132,68 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const
...
+  size_t tramp_size;
+  size_t tramp_npages;
+  uintptr_t dyncode_pnum_child;
+  uintptr_t parent_start_addr;
+  uintptr_t child_start_addr;
+  uintptr_t tramp_pnum;
+
+  /* Convert pathname from user path, set binary */
+  sys_pathname = NaClUserToSysAddr(nap, path);
+  binary = sys_pathname ? strdup(sys_pathname) : 0;
```
jesings: Semantic thing, NULL should be used here instead of 0

jesings: Even if it wasn't that way in the original

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4163,24 +4132,68 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const
...
       env = (uintptr_t)env == kNaClBadAddress ? 0 : env;
-      new_envp[i] = env ? strdup(env) : 0;
       if (!env) {
+        new_envp[i] = 0;
         break;
       }
+      else {
+        int envsize = NACL_ENV_PREFIX_LENGTH + strlen(env) + 1;
+        new_envp[i] = calloc(envsize, sizeof(char));
+        snprintf(new_envp[i], envsize, "%s%s", NACL_ENV_PREFIX, env);
```
jesings, line 4142: snprintf is likely significantly slower than 2 strcpys, but I'll leave it up to you whether we care

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4163,24 +4132,68 @@ int32_t NaClSysExecve(struct NaClAppThread *natp, char const *path, char *const
...
+
+  /* We've already cleaned the native environment, so just supply extra args from this syscall */
+  if (!NaClEnvCleanserInit(&env_cleanser, (char const *const *)new_envp, 0)) {
+    NaClLog(LOG_ERROR, "%s\n", "Failed to initialize environment cleanser");
+    NaClEnvCleanserDtor(&env_cleanser);
+    goto fail;
+  }
+
+  nap->clean_environ = NaClEnvCleanserEnvironment(&env_cleanser);
+  ret = NaClSysExecv(natp, path, argv);
```
jesings, line 4156: You need to free new_envp after execv returns, you currently have a memory leak

jesings (commented):
I'm glad we didn't have to kludge anything new, I don't like the NACLENV_ prefix stuff but it should work well enough.
Overall this should be really helpful to have working!

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: I had it like that initially, I think this is clearer and won't be called enough to really be a performance problem.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: Yup, got it.

rennergade:
Yeah, well the NACLENV_ thing is just extending what already exists from the original NACL implementation, not anything new so I think it's worth sticking with instead of reinventing this.

@jesings I made your changes, think it's good now.

PR #86, opened 2022-01-28
(Partially) Implemented Ioctl

youngfalcon:


rennergade (approved):
Looks good!

kapkic (approved):
LGTM

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1238,70 +1238,38 @@ int32_t NaClSysLseek(struct NaClAppThread *natp,
...
-
-  fd = fd_cage_table[nap->cage_id][d];
+           nap->cage_id, (uintptr_t)natp, d, request,
+           (uintptr_t)arg_ptr);
 
-  /* check for closed fds */
-  if (fd < 0) {
-    retval = -NACL_ABI_EBADF;
-    goto cleanup;
+  sysaddr = NaClUserToSysAddr(nap, (uintptr_t) arg_ptr);
```
jesings, line 1253: I think you need to check whether it's actually supposed to be a pointer type here, i.e. for I_POP, I_FLUSH, etc. we don't want to convert the pointer.

jesings (commented):
Missing one crucial check, otherwise good

jesings:
I also think it's worth mentioning that they layout of the integer being put in the pointer variable is not guaranteed to be the same as that of a union but it should be totally fine on x86.

src/trusted/service_runtime/nacl_syscall_common.c
youngfalcon: The I_POP seems to be an anomaly in requiring a 0, but I think it can be dealt with by simply passing a pointer to an integer 0 rather than restructuring the whole thing. I want to be agnostic about the input type on the GLIBC/NaCl side as much as possible so that minimal change is ever necessary outside SafePosix.

jesings: I_FLUSH however will require an integer, which will break on NaClUserToSysAddr

youngfalcon: Hm, I understand that. But again, we can solve it in glibc the same way (by creating pointers to those values) as we will create checks for the ones we support anyways (these calls are told to be unimplemented directly in GlibC right now so they do not reach this area.). 

Otherwise, if we truly want to pass values for these functions, we will have to explicitly separate different syscall types a lot in all 3 repos and we will have to deal with explicit casts to pointer types or int types.

Either way we go, such changes can be implemented if we add those syscalls as adding the checks won't do anything for now.

jesings: Ah ok. I think as much as possible we should try to avoid changing the way arguments are handled in syscalls. I think at least a comment that this as is will not work for ioctls which take integers would be warranted though, and then this PR will be good to go.

rennergade (approved):
This looks good now!

PR #88, opened 2022-01-31
Socketpair fix

rennergade:
NaClSysSocketpair was sending a user address to RustPOSIX which is incorrect, and not using CopyOutToUser. This creates an array for the syscall and copies that out once it completes.

Please ignore that I named this branch socke**_r_**pair

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -3155,7 +3155,7 @@ int32_t NaClSysSocketPair(struct NaClAppThread *natp,
   struct NaClApp          *nap = natp->nap;
   void                    *hd_struct = NULL;
   struct NaClHostDesc     *hd;
-  int                     lind_fd;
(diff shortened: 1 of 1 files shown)


kapkic (approved):
Quick fix, looks good!

youngfalcon (approved):
LGTM!

jesings (commented):
I would just like to make sure that we try to as much as possible have the failures happen before rustposix (so we don't leave orphaned file descriptors). I think you should check the address range is valid before calling NaClCopyOutToUser perhaps with a NaClUserToSysAddrRange? If not that you could do lind_close on both of the lind fds.
Otherwise this is a great and straightforward fix to a baffling problem.

jesings (approved):
Looks good now!

rennergade:
Yeah, CopyOutToUser checks address range internally, so I just added closes for the RustPOSIX fds on error. Since this is the same as how SysPipe works, I added those checks there as well.

PR #93, opened 2022-02-23
Fixed all the necessary things for gethostbyname/flask

jesings:


jesings:
Unsurprising that it failed this check as we changed the argument order for this recvfrom both here and in glibc and this only has the changes here. With both branches, the tests pass for me.

rennergade (approved):
looks great!

kapkic (approved):
LGTM!

PR #95, opened 2022-03-01
Implemented pipe2

jesings:


jesings:
Ok this failed on building glibc...... I'm rerunning but that is weird.

jesings:
Ok I will have to look into that ENOPERM from pipe.

rennergade:
> Ok I will have to look into that ENOPERM from pipe.

Doesn't it need both branches to run correctly?

jesings:
Oh yeah duh, it calls pipe2 from pipe.

kapkic:
Glibc fail happens randomly, building nacl and glibc usually resolves it.
I'm doing a rerun to make sure it's okay.

jesings:
Yep I just ran the tests again manually with both branches and they were good.

jesings:
Yep the failure on the test cases is expected, we should merge the safeposix one in and then we can rerun.

jesings:
Glibc failed again

PR #98, opened 2022-03-10
Clean FD conversion in syscall_common

rennergade:
PR for https://github.com/Lind-Project/lind_project/issues/209

Adds checks to syscall_common functions that utilize file descriptors to make sure they're positive and below FILEDESC_MAX.


jesings (approved):
Seems like pretty reasonable sanity checking. Is FILE_DESC_MAX really a tight bound? Or should it be >= FILE_DESC_MAX?

youngfalcon (approved):
Looks good to me.

rennergade:
> Seems like pretty reasonable sanity checking. Is FILE_DESC_MAX really a tight bound? Or should it be >= FILE_DESC_MAX?

Good catch, just fixed that.

PR #99, opened 2022-03-11
CLOEXEC fixes

rennergade:
PR from @jesings and I that fixes a few errors with CLOEXEC
- now doesn't pass flags in dup/dup2
- only closes on exec and not fork
- fixes fcntlset for cloexec


kapkic (approved):
LGTM as well!

PR #102, opened 2022-03-23
Socket deadlock, Accept fix, Select fix

rennergade:
This PR fixes:

1. A deadlock issue on I/O calls to read and write when writing to sockets
2. Properly creates an accept nacl desc when accept succeeds
3. Small fix for select that uses calloc instead of malloc for the bitfield

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -5216,7 +5216,6 @@ int32_t NaClSysAccept(struct NaClAppThread *natp,
   ndp = GetDescFromCagetable(nap, sockfd);
   if (!ndp) {
     NaClLog(2, "NaClSysAccept was passed an unrecognized file descriptor, returning %d\n", sockfd);
-    free(nd);
     return -NACL_ABI_EBADF;
   }
   
@@ -5225,7 +5224,6 @@ int32_t NaClSysAccept(struct NaClAppThread *natp,
  
   if ((void*) kNaClBadAddress == syslenaddr) {
     NaClLog(2, "NaClSysAccept could not translate buffer address, returning %d\n", -NACL_ABI_EFAULT);
-    free(nd);
     userfd = -NACL_ABI_EFAULT; // As we return userfd instead of a retvalue, changed ret with userfd.
     goto cleanup;
   }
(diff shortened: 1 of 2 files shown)


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -5185,12 +5213,6 @@ int32_t NaClSysAccept(struct NaClAppThread *natp,
   NaClLog(2, "Cage %d Entered NaClSysAccept(0x%08"NACL_PRIxPTR", %d, 0x%08"NACL_PRIxPTR", 0x%08"NACL_PRIxPTR")\n",
           nap->cage_id, (uintptr_t) natp, sockfd, (uintptr_t) addr, (uintptr_t) addrlen);
 
-  nd = malloc(sizeof(struct NaClHostDesc));
-  if (!nd) {
-    NaClLog(2, "NaClSysAccept could not allocate room for returning NaCl desc\n");
-    return -NACL_ABI_ENOMEM;
-  }
-
```
jesings, line 5193: Now that you have moved the allocation of nd down, you cannot have in every error out below this `free(nd)`. I would be a proponent of moving the allocation back up here because IMO the earlier we can handle an ENOMEM the better.

jesings: The alternative is of course just to remove the frees below this

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -965,7 +981,7 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
   /* Lind - we removed the VMIOWillStart and End functions here, which is fine for Linux
    * See note in sel_ldr.h
    */
-  read_result = ((struct NaClDescVtbl const *)ndp->base.vtbl)->Read(ndp, (void *)sysaddr, count);
+  read_result = lind_read(lindfd, (void *)sysaddr, count, nap->cage_id);
```
jesings, line 984: This is simply beautiful

jesings (commented):
Looks good, one big issue.

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: Id rather allocate after lind_accept since we dont need to make a descriptor if its an error.

rennergade: fixed

youngfalcon (approved):
Looks good to me.

PR #103, opened 2022-03-24
Revert Errno Handling to be passed to libc

rennergade:
This takes out the errno handler we added to lind_platform, and passes that to libc. It also moves mmap errno handling to its HostDesc function.

There is also a small epoll retval fix.

### DIFF
--- a/src/shared/platform/posix/nacl_host_desc.c
+++ b/src/shared/platform/posix/nacl_host_desc.c
@@ -201,6 +201,8 @@ uintptr_t NaClHostDescMap(struct NaClHostDesc *d,
    * start address.
    */
   mapbottom = lind_mmap(start_addr, len, tmp_prot, host_flags, desc, offset, whichcage);
+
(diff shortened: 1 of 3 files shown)


src/shared/platform/posix/nacl_host_desc.c
```diff
@@ -201,6 +201,8 @@ uintptr_t NaClHostDescMap(struct NaClHostDesc *d,
    * start address.
    */
   mapbottom = lind_mmap(start_addr, len, tmp_prot, host_flags, desc, offset, whichcage);
+
+  if ((unsigned) mapbottom > (0xffffffffu - 256)) mapbottom = MAP_FAILED;
   /* MAP_FAILED is -1, so if we get that as our bottom 32 bits, we 
    * return a long -1 as our return value. Otherwise, combine the 
    * top bits and bottom bits into our full return value.
```
jesings, line 213: Update comment

jesings (approved):
Looks good, just change the comment.

PR #104, opened 2022-03-29
Fix HostDesc functions so errno falls through to glibc

rennergade:
These functions were checking if -1 was returned from the old dispatcher format, now they return the negative retval.

### DIFF
--- a/src/shared/platform/posix/nacl_host_desc.c
+++ b/src/shared/platform/posix/nacl_host_desc.c
@@ -326,9 +326,9 @@ int NaClHostDescOpen(struct NaClHostDesc  *d,
           path, posix_flags, mode);
   host_desc = lind_open(path, posix_flags, mode, d->cageid);
   NaClLog(3, "NaClHostDescOpen: got descriptor %d\n", host_desc);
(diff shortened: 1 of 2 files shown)


src/shared/platform/posix/nacl_host_desc.c
```diff
@@ -326,9 +326,9 @@ int NaClHostDescOpen(struct NaClHostDesc  *d,
           path, posix_flags, mode);
   host_desc = lind_open(path, posix_flags, mode, d->cageid);
   NaClLog(3, "NaClHostDescOpen: got descriptor %d\n", host_desc);
-  if (-1 == host_desc) {
-    NaClLog(2, "NaClHostDescOpen: open returned -1, errno %d\n", errno);
-    return -NaClXlateErrno(errno);
+  if (host_desc < 0) {
+    NaClLog(2, "NaClHostDescOpen: open returned -1, errno %d\n", host_desc);
```
jesings: This log should be changed from returned -1 I think

jesings (approved):
Looks great, one nitpick

kapkic (approved):
LGTM

PR #106, opened 2022-04-07
Address Checking (EFAULT)

rennergade:
This PR establishes address checking during user to sys translation in NaCl. This enables us to return EFAULT for non-mapped memory regions instead of letting the system fault.

This PR also:
- includes re-organizing NULL terminated string copying so it won't overflow/and deal with page boundaries.
- Checks that exec pointer arrays are 8-byte aligned
- Implements vmmap entry caching to keep improvements in I/O speed.

Kudos to @jesings for laying the groundwork here many moons ago.


kapkic (approved):
Looks good!

PR #107, opened 2022-04-14
Address Checking (EFAULT)

rennergade:
This PR establishes address checking during user to sys translation in NaCl. This enables us to return EFAULT for non-mapped memory regions instead of letting the system fault.
This PR also:
- includes re-organizing NULL terminated string copying so it won't overflow/and deal with page boundaries.
- Checks that exec pointer arrays are 8-byte aligned
- Implements vmmap entry caching to keep improvements in I/O speed.
- Fixes concurrency issue with nacl fd allocation, cleans up fd_cage_table

Kudos to @jesings for laying the groundwork here many moons ago.


src/trusted/service_runtime/nacl_copy.c
```diff
@@ -51,17 +55,42 @@ int NaClCopyInFromUserZStr(struct NaClApp *nap,
...
-  strncpy(dst_buffer, (char *) src_sys_addr, dst_buffer_bytes);
+  while (1) {
+    uintptr_t page_end = NaClVmmapCheckAddrMapping( &nap->mem_map, check_addr >> NACL_PAGESHIFT, 1, NACL_ABI_PROT_READ);
+    if (!page_end) break;
+    int page_room = (page_end << NACL_PAGESHIFT) - check_addr;
+    int dst_bytes_remaining = dst_buffer_bytes - bytes_copied;
+    int copy_bytes = page_room < dst_bytes_remaining ? page_room : dst_bytes_remaining;
+    strncpy(dst_buffer + bytes_copied, (char *) src_sys_addr + bytes_copied, copy_bytes);
+    bytes_copied = bytes_copied + copy_bytes;
+    if (strnlen(dst_buffer, bytes_copied) < bytes_copied) break;
```
jesings, line 90: Shouldn't we just get the output of strncpy and subtract the base pointer rather than having to go over all the bytes again with a strnlen?

jesings, line 90: Shouldn't we just get the output of strncpy and subtract the base pointer rather than having to go over all the bytes again with a strnlen?

src/trusted/service_runtime/nacl_copy.c
```diff
@@ -35,6 +36,9 @@ int NaClCopyInFromUserAndDropLock(struct NaClApp *nap,
                                   size_t         num_bytes) {
   uintptr_t src_sys_addr;
 
+  /* We don't need to check protections here
+   * this is used by nacl_syscall_handlers, and we can check the actual args later
+   */
```
jesings: You either need to check here or check before, as otherwise the memcpy can fault. So either update the comment or the code.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -765,27 +762,23 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
...
   */
-  
-  retval = NextFd(nap->cage_id);
-  if(hd) {
-    hd->userfd = retval;
+
+  userfd = AllocNextFd(nap, hd);
+  if (userfd < 0) {
+    free(hd);
+    return -NACL_ABI_ENFILE;
```
jesings: Yeah this is what I was talking about at the meeting about what's complicated. The issue here is that it will leave an orphaned file descriptor within rustposix which is why atomically reserving/assigning at the end is a minefield. I think you can probably just do a lind_close if need be there but I don't particularly like that and more thought should be put into potentially handling cases like this.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -5249,19 +5286,21 @@ int32_t NaClSysAccept(struct NaClAppThread *natp,
...
+  hd->cageid = nap->cage_id;
 
-  ret = NaClSetAvail(nap, ((struct NaClDesc *) NaClDescIoDescMake(nd)));
-  userfd = NextFd(nap->cage_id);
-  fd_cage_table[nap->cage_id][userfd] = ret;
+  userfd = AllocNextFd(nap, hd);
+  if (userfd < 0) {
+    free(hd);
+    userfd = -NACL_ABI_ENFILE;
+  }
```
jesings: You still leave an orphaned fd in rustposix here

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -1821,20 +1821,53 @@ void InitializeCage(struct NaClApp *nap, int cage_id) {
...
-int NextFd(int cage_id){
+int CancelFds(struct NaClApp *nap, int userfds[2], int iterations) {
+
+  for (int i = 0; i < iterations; i++) {
+    int naclfd = fd_cage_table[nap->cage_id][userfds[i]];
+    NaClSetDesc(nap, naclfd, NULL);
+    fd_cage_table[nap->cage_id][userfds[i]] = -1;
+  }
+
+}
```
jesings, line 1832: When you cancel the fds, you still leave the orphaned descs in rustposix

src/trusted/service_runtime/sel_mem.c
```diff
@@ -540,9 +542,9 @@ int NaClVmmapEntryMaxProt(struct NaClVmmapEntry *entry) {
 }
 
 int NaClVmmapCheckExistingMapping(struct NaClVmmap  *self,
-                                  uintptr_t         page_num,
-                                  size_t            npages,
-                                  int               prot) {
+                              uintptr_t         page_num,
+                              size_t            npages,
+                              int               prot) {
```
jesings: Why do this deindentation?

src/trusted/service_runtime/sel_mem.c
```diff
@@ -579,6 +581,75 @@ int NaClVmmapCheckExistingMapping(struct NaClVmmap  *self,
...
+    int                     flags = ent->prot;
+   
+    //If the page does not not have PROT_NONE, force the PROT_READ flag
+    //This behavior is unspeicified by POSIX, but Linux acts in this way
+    if((flags & (NACL_ABI_PROT_EXEC | NACL_ABI_PROT_READ | 
+                NACL_ABI_PROT_WRITE)) != PROT_NONE){
+        flags |= PROT_READ;
+    }
+
+    if (ent->page_num <= page_num && region_end_page <= ent_end_page) {
```
jesings, line 614: I think(?) the ent_end_page page is one greater than what we actually want, so do a < there? I'm really not sure of this, so confirm the behavior here

src/trusted/service_runtime/sel_mem.c
```diff
@@ -579,6 +581,75 @@ int NaClVmmapCheckExistingMapping(struct NaClVmmap  *self,
...
+    int                     flags = ent->prot;
+    
+    //If the page does not not have PROT_NONE, force the PROT_READ flag
+    //This behavior is unspeicified by POSIX, but Linux acts in this way
+    if((flags & (NACL_ABI_PROT_EXEC | NACL_ABI_PROT_READ | 
+                NACL_ABI_PROT_WRITE)) != PROT_NONE){
+        flags |= PROT_READ;
+    }
+
+    if (ent->page_num <= page_num && region_end_page <= ent_end_page) {
```
jesings, line 634: Same here, I know I wrote this 2 years ago but please check this.

jesings (changes requested):
In general great progress.
However you do leave a lot of orphaned rustposix fds in cases where you fail, because you error out too late. This could be handled by explicitly doing a lind_close or something like that, but I'm not a fan of that solution (even if we had it before for socketpair in one place that we had to, I'm just kind of resigned to the fact that socketpair is messy and gross)
If there's any way to reserve something without totally rewriting all the desc handling (say by using a dummy value of -2 in the fd_cage_table and a dummy nacldesc or somethign), I would really want to see about that. If that's not possible, then you do have to do the lind_close solution.

src/trusted/service_runtime/sel_ldr.c
rennergade: This actually does call lind_close through the inner DescUnref

src/trusted/service_runtime/sel_mem.c
rennergade: It's not, this is the same as the check in the original function.

src/trusted/service_runtime/sel_mem.c
rennergade: Ha. You actually didn't write this part, so you can worry less!

src/trusted/service_runtime/nacl_copy.c
rennergade: strncpy returns the original dest ptr so this won't work.

rennergade:
@jesings I believe I addressed your concerns in the last commits.

jesings (approved):
Just change the stack comment to say that it has to have been writable in lind_glibc so it must be readable here

PR #108, opened 2022-04-19
Threads - Untrusted Teardown + NapDtor

rennergade:
This PR combines the teardown handler (which was already looked over mostly) and functionality to properly free the NaClApp structs.

I've removed the signal teardown for trusted code since we should really abort if anything happens there now that we are checking vmmap protections for wonky addresses.

The napdtor includes switching to shutting down desc's through NaCl on exit, and handling situations where a cage becomes an orphan, as well as creating the Dtor function which is triggerd in NaClAppThreadTeardown.


src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -208,6 +217,34 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap, int child_cage_id, enum Na
   return nap_child;
 }
 
+void NaClAppCloseFDs(struct NaClApp *nap) {
+  struct NaClDesc *ndp = NULL;
+  int             fd = 0;
+  int             i;
```
jesings: There's no reason not to declare i within the while loop given that it's a loop variable and it's not used after the loop

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -208,6 +217,34 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap, int child_cage_id, enum Na
   return nap_child;
 }
 
+void NaClAppCloseFDs(struct NaClApp *nap) {
+  struct NaClDesc *ndp = NULL;
+  int             fd = 0;
```
jesings: This assignment to fd is never used, as it's reassigned at the start of the loop

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -208,6 +217,34 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap, int child_cage_id, enum Na
   return nap_child;
 }
 
+void NaClAppCloseFDs(struct NaClApp *nap) {
+  struct NaClDesc *ndp = NULL;
```
jesings: Ndp can be declared at the top of the innermost scope it will be used in, but that's just preference.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -364,14 +391,45 @@ void NaClAppThreadTeardown(struct NaClAppThread *natp) {
     else {
       NaClLog(1, "[NaClAppThreadTeardown][parent %d] removed cage: cage_id = %d\n", nap_parent->cage_id, nap->cage_id);
     }
+
     NaClXCondVarBroadcast(&nap_parent->children_cv);
     NaClXMutexUnlock(&nap_parent->children_mu);
+  }
 
+  /* Sadly we've created orphans :'(
+   * Remove parent from any children, and hope they don't become the Batman
```
jesings, line 387: Haha

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -364,14 +391,45 @@ void NaClAppThreadTeardown(struct NaClAppThread *natp) {
...
+   */
+  NaClXMutexLock(&nap->children_mu);
+  for (int i = 0; i < (&nap->children)->ptr_array_space; i++) {
```
jesings: Isn't ptr_array_space not the right field (i.e. the number of slots allocated rather than the number of slots used?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -653,6 +744,17 @@ int NaClAppThreadSpawn(struct NaClAppThread     *natp_parent,
...
 
+
+  if (!cage_thread) {
+    natp_child->is_cage_mainthread = true;
+    natp_child->cage_mainthread = NULL;
+    natp_child->tearing_down = false;
+  } 
+  else {
+    natp_child->is_cage_mainthread = false;
+    natp_child->cage_mainthread = natp_parent;
```
jesings: Wait is this always true?
I'd think you should probably just have the natp for the cage mainthread refer to itself with the cage_mainthread field and then you can set `natp_child->cage_mainthread = natp_parent->cage_mainthread` here.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -710,6 +812,113 @@ void NaClAppThreadDelete(struct NaClAppThread *natp) {
...
 }
 
+
+/**
+ * The following functions are used to reap any cages which have received a fatal signal.
+ * On launch, sel_main creates the Reaper thread which calls the fault teardown functions when
+ * a faulted thread is caught in the signal handler (nacl_signal.c)
+ * 
+ * The fault teardown function will cycle through all child threads in cage,
+ * tear them down, and call pthread_cancle upon them
```
jesings: Typo

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -710,6 +812,113 @@ void NaClAppThreadDelete(struct NaClAppThread *natp) {
...
+
+void AddToFatalThreadTeardown(struct NaClAppThread *natp) {
+    if (natp_to_teardown == natp) return;
+    NaClXMutexLock(&teardown_mutex);
+    natp_to_teardown = natp;
+    natp->tearing_down = true;
+    NaClXCondVarBroadcast(&reapercv);
+    NaClXMutexUnlock(&teardown_mutex);
+}
+
```
jesings: Hmm how come this doesn't just let anyone clobber over the values in this field whenever they want after they're both assigned?

src/trusted/service_runtime/nacl_signal_common.c
```diff
@@ -134,7 +138,23 @@ void NaClSignalHandleUntrusted(int signal,
...
+    NaClTlsSetCurrentThread(NULL);
+    // signal for teardown
+    if (natp->is_cage_mainthread) {
+      if (!natp->tearing_down) AddToFatalThreadTeardown(natp);
+      NaClUntrustedThreadSuspend(natp, 0);   
+    } else {
+      if (!natp->cage_mainthread->tearing_down) AddToFatalThreadTeardown(natp->cage_mainthread);
+    }
+    // wait here while we get cleaned up
+    while (1);
```
jesings, line 156: At this point you should probably do a `pause()` instead of busy loop?

jesings: That is the purpose of pause after all

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -310,44 +318,83 @@ int NaClAppWithSyscallTableCtor(struct NaClApp               *nap,
...
- cleanup_none:
+  cleanup_none:
   return 0;
 }
 
 int NaClAppCtor(struct NaClApp *nap) {
   return NaClAppWithSyscallTableCtor(nap, nacl_syscall);
 }
 
+void NaClAppDtor(struct NaClApp *nap) {
```
jesings, line 362: This is much better than whatever garbage we were doing before.

jesings (commented):
I'm not sure what happens if the main thread calls pthread_exit but there are still other threads in the process.
Overall this is awesome stuff.

src/trusted/service_runtime/nacl_signal_common.c
rennergade, line 156: I'm almost certain I can't use pause() in a signal handler. My capabilities here are pretty minimal.

src/trusted/service_runtime/nacl_app_thread.c
rennergade: There isn't really functionality to just go over the slots used so I have to run over all that are allocated. This is still a small number.

jesings: The num_entries field should do just that though?

src/trusted/service_runtime/nacl_signal_common.c
jesings: Oh right makes sense

rennergade:
Going to close this for a bit while I fix it up.

rennergade:
@jesings I think I addressed your concerns here.

rennergade:
fork_simple is being wonky in the test suite, but @kapkic is fixing that.

rennergade:
Dealing with that pesky fork_simple test actually exposed a bug in how we're duplicating nap information on fork/exec. Closing this until I fix that bug.

rennergade:
Finally got that green check mark.

I think this is good to go now. We changed the failing test (fork_simple) to a test that isn't insane. But regardless I fixed what was breaking the original one, which was a problem with the service threads once I started freeing the Nap structs properly.

I removed the service threads entirely, we don't use them at all. So this now makes things a bit cleaner to look at in gdb as a positive side effect (as well as being faster? probably negligibly but still...)

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -65,9 +73,14 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap, int child_cage_id, enum Na
...
   nap_child->argc = nap_parent->argc;
-  nap_child->argv = nap_parent->argv;
-  nap_child->binary = nap_parent->binary;
-  nap_child->nacl_file = nap_parent->nacl_file ? nap_parent->nacl_file : LD_FILE;
+
+  if (tl_type == THREAD_LAUNCH_FORK) {
+    nap_child->argv = calloc((nap_child->argc + 1), sizeof(char*));
+    for (int i = 0; i < nap_child->argc; i++) nap_child->argv[i] = strdup(nap_parent->argv[i]);
+    nap_child->binary = strdup(nap_parent->binary);
+  }
```
jesings: I'd rather this part be after all of the other nap_child assigns just to make the function better organized.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -147,14 +159,8 @@ struct NaClApp *NaClChildNapCtor(struct NaClApp *nap, int child_cage_id, enum Na
   if (!nap_child->validator->FixCPUFeatures(nap_child->cpu_features)) {
     NaClLog(LOG_FATAL, "This CPU lacks features required by fixed-function CPU mode.\n");
   }
-  if (!NaClAppLaunchServiceThreads(nap_child)) {
-    NaClLog(LOG_FATAL, "Launch service threads failed\n");
-  }
-
   
   /* duplicate file descriptor table starting at child_fd = 3 (0-2 setup previously)*/
-  NaClXMutexLock(&nap_parent->mu);
```
jesings, line 156: Do we not want to lock this to make sure there are no race conditions with the parent's fd_cage_table?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -398,42 +454,71 @@ void NaClAppThreadTeardown(struct NaClAppThread *natp) {
...
   NaClRemoveThreadMu(nap, natp->thread_num);
   NaClLog(3, " unlocking thread\n");
   NaClXMutexUnlock(&natp->mu);
-  NaClLog(3, " unlocking thread table\n");
-  NaClXMutexUnlock(&nap->threads_mu);
-  NaClLog(3, " unregistering signal stack\n");
-  NaClSignalStackUnregister();
+
+  // again we trigger the lock outside for non-active threads, we also cant unregister the sig stack
+  // here so lets destroy everything
```
jesings: This comment doesn't make sense right before you unregister the signal stack...

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -398,42 +454,71 @@ void NaClAppThreadTeardown(struct NaClAppThread *natp) {
...
+  if (active_thread) {
+    NaClLog(3, " unlocking thread table\n");
+    NaClXMutexUnlock(&nap->threads_mu);
+    NaClLog(3, " unregistering signal stack\n");
+    NaClSignalStackUnregister();
+  }
+
+
+  if (natp->is_cage_mainthread) {
+    // we have to wait for NaClWaitForMainThreadToExit on Main/Exec
```
jesings: Why for exec? (not sure I understand 100% what's being waited on).

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -398,42 +454,71 @@ void NaClAppThreadTeardown(struct NaClAppThread *natp) {
...
   NaClLog(3, " freeing thread object\n");
   NaClAppThreadDelete(natp);
-  NaClLog(3, " NaClThreadExit\n");
+  
+  // if were handling threads, we'll call pthread_cancel from outside, otherwise lets leave
+  if (active_thread) {
+    NaClLog(3, " NaClThreadExit\n");
+    NaClThreadExit();
+    NaClLog(LOG_FATAL, "NaClAppThreadTeardown: NaClThreadExit() should not return\n");
+    /* NOTREACHED */
```
jesings: I don't think this comment is necessary

jesings: Even if it was there previously.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -710,6 +806,114 @@ void NaClAppThreadDelete(struct NaClAppThread *natp) {
...
+}
+
+void DestroyFatalThreadTeardown(void) {
+  NaClMutexDtor(&teardown_mutex);
+  NaClMutexDtor(&reapermut);
+  NaClCondVarDtor(&reapercv);
+}
+
+
+void AddToFatalThreadTeardown(struct NaClAppThread *natp) {
```
jesings, line 842: This mutex locking architecture still seems highly problematic to me given that it's undefined behavior for a thread to unlock a mutex that was locked by another thread.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -710,6 +806,114 @@ void NaClAppThreadDelete(struct NaClAppThread *natp) {
...
+    
+    lind_exit(status, nap->cage_id);
+    (void) NaClReportExitStatus(nap, NACL_ABI_W_EXITCODE(status, 0));
+    thread = &natp_to_teardown->host_thread;
+    NaClXMutexUnlock(&nap->threads_mu);
+
+    NaClAppThreadTeardownInner(natp_to_teardown, false);
+    NaClThreadCancel(thread);
+    natp_to_teardown = NULL;
+    NaClXMutexUnlock(&teardown_mutex);
```
jesings: weird indentation

src/trusted/service_runtime/nacl_app_thread.c
jesings: From what I understand, this function is just called by any old thread, and the reaper thread is responsible for unlocking in FatalThreadTeardown.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -710,6 +806,114 @@ void NaClAppThreadDelete(struct NaClAppThread *natp) {
...
+
+void ThreadReaper(void* arg) {
+  while (reap) {
+    NaClXMutexLock(&reapermut);
+
+    NaClXCondVarWait(&reapercv, &reapermut);
+    if (reap) FatalThreadTeardown();
+
+    NaClXMutexUnlock(&reapermut);
+  }
```
jesings: Is there a reason to have the locking/unlocking in the loop, rather than outside? You get a signal when reap is set to false, so this wouldn't change anything really.

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -710,6 +806,114 @@ void NaClAppThreadDelete(struct NaClAppThread *natp) {
...
+void LaunchThreadReaper(void) {
+  InitFatalThreadTeardown();
+  if (!NaClThreadCtor(&reaper, ThreadReaper, NULL, NACL_KERN_STACK_SIZE)) {
+    NaClLog(LOG_FATAL, "%s\n", "Failed to initialize reaper");
+  }
+}
+
+void DestroyReaper(void) {
+  reap = false;
+  NaClXCondVarBroadcast(&reapercv);
```
jesings: For a single thread, that can recieve a signal on a cv, you should use signal instead of broadcast.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -381,13 +381,20 @@ int32_t NaClSysGetppid(struct NaClAppThread *natp) {
 
 int32_t NaClSysExit(struct NaClAppThread  *natp,
                     int                   status) {
-
   struct NaClApp *nap = natp->nap;
+
+  /* to close a cage we need to unref the vmmap before fofficially
```
jesings: Typo

jesings (changes requested):
Sorry, I still think the mutex thing is not quite right. Overall this is looking much better though.

src/trusted/service_runtime/nacl_app_thread.c
rennergade: I moved this so the mutex is actually held outside of the entire ChildCtor function.

src/trusted/service_runtime/nacl_app_thread.c
rennergade: I forgot that was the case. Makes sense. Will replace with a CV.

rennergade:
Alright, I think this is good now. @jesings I made the changes you requested and changed the arg translation a bit for fork/exec so they could be freed appropriately.

jesings (approved):
Looks good!

PR #110, opened 2022-05-12
Chmod/Rename

rennergade:
Hooks in chmod and rename to the implementation in SafePOSIX

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -1573,6 +1573,32 @@ cleanup:
   return retval;
 }
 
+int32_t NaClSysChmod(struct NaClAppThread *natp,
+                     uint32_t             pathname,
+                     int                  mode) {
+  struct NaClApp *nap = natp->nap;
+  char           path[NACL_CONFIG_PATH_MAX];
+  int32_t        retval = -NACL_ABI_EINVAL;
+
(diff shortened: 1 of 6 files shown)


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1573,6 +1573,32 @@ int32_t NaClSysChdir(struct NaClAppThread *natp,
...
+  char           path[NACL_CONFIG_PATH_MAX];
+  int32_t        retval = -NACL_ABI_EINVAL;
+
+  NaClLog(2, "Cage %d Entered NaClSysChmod(0x%08"NACL_PRIxPTR", "
+          "%d, %d)\n", nap->cage_id, (uintptr_t) natp, pathname, mode);
+
+  if (!NaClAclBypassChecks) {
+    retval = -NACL_ABI_EACCES;
+    goto cleanup;
+  }
```
jesings, line 1589: What is the purpose of this? comment it

jesings (commented):
Overall looks good, I did find one thing that needs more explaining

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: This is in every file/directory related syscall mkdir/chdir/rmdir/getcwd etc... Just followed that to be consistent. Seems like an unesscesary remnant but don't really want to screw with the whole thing right now and keep things consistent.

PR #111, opened 2022-05-13
libc 2.34 fixes

rennergade:
Some minor fixes so that we can compile with native libc > 2.34 which us allows us to update to newer tools in our containers (perf, RR, etc...)

These revolve around libc changing how the macros work for PTHREAD_STACK_MIN and SIGSTKSZ (https://lists.gnu.org/archive/html/info-gnu/2021-08/msg00001.html). We can get around this by undefining `_GNU_SOURCE` in the nacl threads and signal stack files. We also have to manually define the SIGSTKSZ value in one of the nacl tests because of a dependency knot with signal and ucontext macros, but this is not really a big deal.

### DIFF
--- a/src/shared/platform/posix/nacl_threads.c
+++ b/src/shared/platform/posix/nacl_threads.c
@@ -7,6 +7,7 @@
 /*
  * NaCl Server Runtime threads implementation layer.
  */
+#undef _GNU_SOURCE
 
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


src/shared/platform/posix/nacl_threads.c
```diff
@@ -18,6 +19,7 @@
 #include <sys/mman.h>
 #include <sys/types.h>
 #include <signal.h>
+
```
jesings: Weird space addition

jesings (approved):
Not the cleanest, but looks fine

PR #112, opened 2022-05-19
Fork Mutex Fix

rennergade:
This PR fixes a double mutex unlock that somehow has been in the code since before I even started here. Not sure how its remained hidden so long.


kapkic (approved):
Easy fix! LGTM

jesings (approved):
Good. I think the double unlock just ends up having the second unlock fail which is why it's not catastrophic but it's much better not to have it at all

rennergade:
> Good. I think the double unlock just ends up having the second unlock fail which is why it's not catastrophic but it's much better not to have it at all

It started generating some errors, which may be due to the libc upgrade? Either way, glad its fixed.

PR #113, opened 2022-05-23
Select FD Set fix

rennergade:
The NaCl fd set translators didnt skip if the fd wasn't in the cage table, this PR fixes that.


jesings (commented):
What circumstances does this even happen in

rennergade:
The one I saw was with a closed fd 0, presumably after some dup in the program.

jesings:
Ah ok looks good then

PR #114, opened 2022-05-27
WNOHANG

rennergade:
Very simple PR adding WNOHANG to waitpid so bash can do bg processes.


jesings (commented):
It should be `if(options & WNOHANG)`

rennergade:
> 

Oops, of course.

PR #115, opened 2022-05-31
added null argv[0] exception for exec syscall

Devansh-Patel:
Added check for Null argument being passed to exec system call.

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -4769,6 +4769,9 @@ int32_t NaClSysExecv(struct NaClAppThread *natp, char const *path, char *const *
   for (int i = 0; i < new_argc; i++) {
     child_argv[i + 3] = new_argv[i] ? strdup(new_argv[i]) : NULL;
   }
+  if (child_argv[3] == NULL){
+    NaClLog(LOG_FATAL, "%s\n", "A NULL argv[0] was passed through an exec system call.");
+  }  
 
   /* initialize child from parent state */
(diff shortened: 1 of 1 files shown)


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4769,6 +4769,9 @@ int32_t NaClSysExecv(struct NaClAppThread *natp, char const *path, char *const *
   for (int i = 0; i < new_argc; i++) {
     child_argv[i + 3] = new_argv[i] ? strdup(new_argv[i]) : NULL;
   }
+  if (child_argv[3] == NULL){
+    NaClLog(LOG_FATAL, "%s\n", "A NULL argv[0] was passed through an exec system call.");
```
rennergade, line 4773: is this the same log message as native?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4769,6 +4769,9 @@ int32_t NaClSysExecv(struct NaClAppThread *natp, char const *path, char *const *
   for (int i = 0; i < new_argc; i++) {
     child_argv[i + 3] = new_argv[i] ? strdup(new_argv[i]) : NULL;
   }
```
rennergade, line 4771: I think a newline between the for loop and if statement is clearer stylistically to separate them.

src/trusted/service_runtime/nacl_syscall_common.c
Devansh-Patel: Yes, I used the same log message as the native gcc binary.

PR #116, opened 2022-06-07
Zombies

rennergade:
Adds a zombie list for all processes to hold exited children. Removes zombies when waited on.

Generally fixes up comments for waitpid() as well.


kapkic:
LGTM

PR #119, opened 2022-06-27
NDP error check, read/write

rennergade:
Quick fix: Returns EBADF in read()/write() if the nacl desc found is NULL.


justyoyo123 (approved):
Looks good to me.

PR #122, opened 2022-08-02
Epoll Fixes in NaCl, updated epoll_wait

justyoyo123:


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -106,7 +106,7 @@ struct NaClDesc *GetDescFromCagetable(struct NaClApp *nap, int fd) {
     return NULL;
   }
 
-  int naclfd = fd_cage_table[nap->cage_id][fd];
+  int naclfd = fd_cage_table[nap->cage_id][fd]; // search fd_cage_table using using user_fd and cage_id
```
rennergade: We don't need to add this comment.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -6037,54 +6037,37 @@ int32_t NaClSysEpollWait(struct NaClAppThread  *natp, int epfd, struct epoll_eve
...
-          hfd = ((struct NaClDescIoDesc *)ndp)->hd->d;
-          if(pfds[i].data.fd == hfd) {
-              pfds[i].data.fd = j;
-          }
-          NaClDescUnref(ndp);
-      }
-  }
-  
-  NaClFastMutexUnlock(&nap->desc_mu);
+  nfds = lind_epoll_wait(lindepfd, sys_event_array, maxevents, timeout, nap->cage_id);
```
rennergade, line 6064: I think after here we need a longform comment about how there is an fd stored inside the epoll_event, but we leave it as an untranslated userfd, so we don't have to translate it back after the lind call.

We should also comment in epoll_ctl that this is happening.

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -6037,54 +6037,37 @@ int32_t NaClSysEpollWait(struct NaClAppThread  *natp, int epfd, struct epoll_eve
...
   int retval = 0;
   int nfds;
-  int hfd;
-  struct epoll_event *pfds;
-  struct NaClDesc * ndp;
+  int lind_fd;
+  int lindepfd;
+  struct epoll_event *sys_event_array;
+  struct NaClDesc * epndp;
+  struct NaClDesc * eventndp;
```
rennergade: I think we dont need this or the lind_fd if we get rid of the post-translation.

rennergade (changes requested):

rennergade (approved):
looks good!

PR #125, opened 2022-08-30
Modified fork to preserve shared memory mappings in child

jesings:


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -3029,14 +3041,17 @@ int32_t NaClSysShmat(struct NaClAppThread  *natp,
   }
 
   if (length > 0) {
+    //we hack the file_size field to store the shmid, because FILE_DESC_MAX is less than 4096
```
rennergade: So Id much rather add an entry struct field than to hack the length here. Even if its neat and works, seems like it will be really confusing down the road.

rennergade (changes requested):
A lot of good stuff here. I would like the way youre dealing with shmids in the vmentrys changed, and generally more comments in the code and in the actual PR posting.

rennergade (approved):
Awesome. This looks great, thanks for changing how shmid is structured.

PR #127, opened 2022-09-12
Moved mutexes into rustposix and made it nicer in general

jesings:


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4055,28 +3990,15 @@ int32_t NaClSysCondTimedWaitAbs(struct NaClAppThread     *natp,
...
+  retval = lind_cond_timedwait(cond_handle, mutex_handle, (struct timespec*) &trusted_ts, nap->cage_id);
 
-  retval = ((struct NaClDescVtbl const *)cv_desc->base.vtbl)->TimedWaitAbs(cv_desc, mutex_desc, &trusted_ts);
-  NaClDescUnref(cv_desc);
-  NaClDescUnref(mutex_desc);
 cleanup:
   return retval;
 }
 
+//TODO: semaphores currently do not have a proper implementation as the desc numbers cannot be
```
rennergade, line 3999: You mentioned that these aren't reachable because they're not implemented in our glibc, is that correct? If so add that here.

rennergade (approved):
Small comment suggestion but looks good.

Devansh-Patel (approved):
I've reviewed this PR to the best of my knowledge and it looks good to me.

PR #128, opened 2022-09-14
Added in truncate and ftruncate system calls

jesings:


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -5548,6 +5548,40 @@ int32_t NaClSysAccess(struct NaClAppThread *natp,
   return ret;
 }
 
+int32_t NaClSysTruncate(struct NaClAppThread *natp,
+                        uint32_t file, int length) {
```
rennergade, line 5552: this is the path right? other calls just use char*, shouldnt that work?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -5548,6 +5548,40 @@ int32_t NaClSysAccess(struct NaClAppThread *natp,
...
+
+  ret = lind_truncate(path, length, nap->cage_id);
+
+  return ret;
+}
+
+int32_t NaClSysFtruncate(struct NaClAppThread *natp,
+                         int fd, int length) {
+  int32_t ret;
+  struct NaClApp *nap = natp->nap;
```
rennergade, line 5575: You never translate the fd from userfd to lind fd, which should be wrong. May have worked in some corner cases?

rennergade (changes requested):

src/trusted/service_runtime/nacl_syscall_common.c
jesings: There's a comment in syscall_handlers_gen.py that says you should use a uint32_t but we've mostly been ignoring that. I think it makes more sense to follow that because we're really only passing 32 bits of stuff into NaCl here.

src/trusted/service_runtime/nacl_syscall_common.c
jesings: Yeah I think postgres was using truncate and it worked in the corner case of my test... I'll fix it

justyoyo123 (approved):
This pr that adds truncate and ftruncate into NaCl looks good to me.

PR #129, opened 2022-09-19
Fix waitpid

rennergade:
Fixes the WAIT_ANY section of waitpid so that it only checks finished processes (zombies) instead of super-naively looping over all cages. This also fixes a bug for backgrounding where it was getting stuck waiting on a back grounded process instead of returning a finished later one.

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -4901,84 +4901,90 @@ int32_t NaClSysWaitpid(struct NaClAppThread *natp,
   NaClLog(1, "%s\n", "[NaClSysWaitpid] entered waitpid!");
 
   CHECK(nap->num_children < NACL_THREAD_MAX);
-  if (stat_loc_ptr) {
-    *stat_loc_ptr = 0;
-  }
+  if (stat_loc_ptr) *stat_loc_ptr = 0;
 
   // First check if we have children, if not check zombies, if no zombies return ECHILD
-  if (!nap->num_children || pid > pid_max) {
-    struct NaClZombie* zombie = NaClCheckZombies(nap);
-    if (zombie == NULL) ret = -NACL_ABI_ECHILD;
-    else {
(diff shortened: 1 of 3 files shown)


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4933,51 +4939,45 @@ int32_t NaClSysWaitpid(struct NaClAppThread *natp,
...
-      for (int cage_id = 0; cage_id < fork_num; cage_id++) {
+      /* make sure children exist, if not send ABI_ECHILD */
+      if (!nap->num_children && !nap->zombies.num_entries) {
+        ret = -NACL_ABI_ECHILD;
+        NaClXCondVarBroadcast(&nap->children_cv);
+        NaClXMutexUnlock(&nap->children_mu);
+        goto out;
+      }
+
+      NaClXMutexLock(&nap->children_mu);
```
jesings: Wait how are we unlocking it if it's not locked yet?

jesings: Remove the lock and broadcast

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4933,51 +4939,45 @@ int32_t NaClSysWaitpid(struct NaClAppThread *natp,
...
+      struct NaClZombie* zombie = NaClCheckZombies(nap);
+      if (zombie) {
+        if (stat_loc_ptr) *stat_loc_ptr = zombie->exit_status; // set the status pointer if it exists
+        ret = zombie->cage_id;
+        NaClRemoveZombie(nap, ret); //remove from zombie list
 
+        NaClXCondVarBroadcast(&nap->children_cv);
         NaClXMutexUnlock(&nap->children_mu);
+
+        goto out;
```
jesings: You could potentially break instead of goto, but I don't think that's a big deal

jesings (commented):
Mostly looks good

PR #130, opened 2022-09-20
Waitpid Overhaul

rennergade:
This refactors and fixes the entirety of waitpid. This started with fixing a backgrounding bug, but it revealed bugs elsewhere in waitpid when the pid is specified (it would formerly return if any cage had finished.

This refactors all of this to rely on the zombies list, which holds any cage that has exited. It also makes the mutex/CV scheme much more straightforward.

### DIFF
--- a/src/trusted/service_runtime/nacl_syscall_common.c
+++ b/src/trusted/service_runtime/nacl_syscall_common.c
@@ -4878,6 +4878,12 @@ fail:
 #define WAIT_ANY (-1)
 #define WAIT_ANY_PG 0
 
+/*
+ * Wait/Waitpid is used to reap a process's exited children, referred to as zombies
+ * We use the NaClCheckZombies/NaClAddZombies/NaClRemoveZombies functions from sel_ldr.c to manage these zombies
+ * Zombies are added to a parents zombie list when a child exits, in the NaClReportExitStatus function in sel_ldr_standard.c
+ */
+
 int32_t NaClSysWaitpid(struct NaClAppThread *natp,
                        int pid,
                        uint32_t *stat_loc,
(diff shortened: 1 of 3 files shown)


src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -4891,98 +4897,68 @@ int32_t NaClSysWaitpid(struct NaClAppThread *natp,
...
 
   // if WAIT_ANY, we'll busy loop on all children, except in the case of WNOHANG where we only loop once
   if (pid <= 0) {
-    while(1){
 
-      /* Cycle through possible cages up to the fork number (max amount of created cages) */
-      for (int cage_id = 0; cage_id < fork_num; cage_id++) {
+    // check the zombies dynarray, and lazily return the first exited process if it exists
+    zombie = NaClCheckZombies(nap);
+    if (!zombie && (options & WNOHANG)) goto out; // exit here if WNOHANG
```
jesings: I would make a comment that ret is 0 here as it is supposed to be

src/trusted/service_runtime/sel_ldr.c
```diff
@@ -1990,6 +1990,15 @@ struct NaClZombie* NaClCheckZombies(struct NaClApp *nap) {
   return zombie;
 }
 
+struct NaClZombie* NaClCheckZombieById(struct NaClApp *nap, int cage_id) {
+  NaClXMutexLock(&nap->zombie_mu);
+  struct NaClZombie* zombie = NULL;
+  zombie = DynArrayGet(&nap->zombies, cage_id);
```
jesings: No reason to null initialize it and then assign it on the next line

jesings (commented):
Looks good to me. I assume that the cv is broadcast on exit/exec?

Devansh-Patel (approved):
I've reviewed this pull request and it looks good to me.

PR #131, opened 2022-11-04
Fix condvar rpc stubs

rennergade:
This is a small fix for the condvar RPC functions that pairs with @jesings PR fixing mutexes/condvars more generally.


jesings (approved):
Simple and good fix!

AlpacaMax (commented):
Looks good to me

PR #132, opened 2023-01-20
Remove NaCl Descs

rennergade:
This PR finally completes the long awaited task of removing the NaClDesc system from system call pathways. We achieve this by

1. Reverting the opening/loading of runnable-ld.so from interfacing with RustPOSIX. This isn't done by a syscall and I believe should be done like this regardless. This allows us to not interfere with the Desc system for the loader. This also required reverting the HostDesc calls to their original state.
2. Remove Descs from all system call pathways ie now NaClSysDup directly calls lind_dup
3. Update mmap slightly to not use descs. I've also removed the unused pathway for mapping executable code.
4. Update the VMap so that we can store a Max Protection field, This is either RDWR, or the max protection an fd is opened as (ie if a file is opened as RDONLY the max prot is PROT_READ.
5. I simplified how stat buffers worked, essentially removing the middle man that nacl was using.


rennergade:
This is going to fail the CI since its a major revamp and dependent on the other two PRs.

src/trusted/desc/nacl_desc_io.c
```diff
@@ -63,20 +63,14 @@ int NaClDescIoDescCtor(struct NaClDescIoDesc  *self,
...
   if (!NaClDescCtor(basep)) {
     return 0;
   }
   rv = NaClDescIoDescSubclassCtor(self, hd);
   if (!rv) {
     (*NACL_VTBL(NaClRefCount, basep)->Dtor)((struct NaClRefCount *) basep);
   }
-  if (hd) {
-    NACL_VTBL(NaClDesc, basep)->SetFlags(basep, hd->flags & NACL_ABI_O_ACCMODE);
-  }
```
jesings, line 79: Why can we get rid of this?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -98,88 +97,78 @@
...
+  uintptr_t usr_new_last_data_page;
+  uintptr_t last_internal_data_addr;
+  uintptr_t last_internal_page;
+  uintptr_t start_new_region;
+  uintptr_t region_size;
 
   break_addr = nap->break_addr;
 
-  NaClLog(3, "Entered NaClSysBrk(new_break 0x%08"NACL_PRIxPTR")\n",
+  NaClLog(3, "Entered NaClSysBrk(new_break 0x%08" NACL_PRIxPTR ")\n",
```
jesings: You do weird stuff with the spacing in this file, which I think breaks convention with the rest of NaCl. I'm not sure that this is worth fixing, but avoiding it in the future for sweeping single file changes like this might be good

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -98,88 +97,78 @@
...
           new_break);
 
   sys_new_break = NaClUserToSysAddr(nap, new_break);
-  NaClLog(3, "sys_new_break 0x%08"NACL_PRIxPTR"\n", sys_new_break);
+  NaClLog(3, "sys_new_break 0x%08" NACL_PRIxPTR "\n", sys_new_break);
 
-  if (kNaClBadAddress == sys_new_break) {
+  if (kNaClBadAddress == sys_new_break)
+  {
```
jesings: Particularly doing this with the curly braces, breaking K&R style, adds a lot of vertical space to the file that the style used elsewhere does not.

jesings: In general, I can't give this file a good review because there are too many style changes and spacing changes. I suspect that you ran an autoformatter on this, but it really does clutter up the diff to the point of being a lot harder to work on. If you did want to change this many stylistic things, it should have been done in a different PR. Could you extricate these if they are in a single commit?

src/trusted/service_runtime/sel_addrspace.c
```diff
@@ -186,6 +188,7 @@ NaClErrorCode NaClMemoryProtection(struct NaClApp *nap) {
                  NaClSysToUser(nap, start_addr) >> NACL_PAGESHIFT,
                  region_size >> NACL_PAGESHIFT,
                  NACL_ABI_PROT_NONE,
+                 NACL_ABI_PROT_NONE,
```
jesings, line 191: Why do we sometimes use the NaCl ABI constants and sometimes the host ABI constants?
We should be consistent and probably use the NaCl ones.

src/trusted/service_runtime/sel_mem.h
```diff
@@ -20,7 +20,8 @@ EXTERN_C_BEGIN
 
 struct NaClDesc;
 
-#define NACL_MAP_COPY   0x100
+#define NACL_MAP_COPY   0x1003d
```
jesings: Why is this?

src/trusted/service_runtime/sel_mem.h
```diff
@@ -98,7 +101,8 @@ void  NaClVmmapAdd(struct NaClVmmap   *self,
 void  NaClVmmapAddWithOverwrite(struct NaClVmmap  *self,
                                 uintptr_t         page_num,
                                 size_t            npages,
-                                int               prot,
+                                int                prot,
+                                int                maxprot,
```
jesings: fix spacing

jesings (changes requested):
This looks great so far, but nacl_syscall_common.c is so cluttered with formatting changes that appear to have been written by an autoformatter so that I can't really parse easily at all what changes are significant and what aren't. See my comment for more about that. I would really appreciate a version of this PR that has diffs less cluttered by unrelated spacing changes, and then I would be able to give nacl_syscall_common.c a proper review in the scope of this PR.

src/trusted/service_runtime/sel_mem.h
rennergade: no idea how that slipped in, nice catch

src/trusted/service_runtime/sel_addrspace.c
rennergade: If you're referring to like "PROT_RX" thats actually a nacl macro

```#define PROT_RX (NACL_ABI_PROT_READ | NACL_ABI_PROT_EXEC)```

Comment (jesings) [reply src/trusted/service_runtime/sel_addrspace.c:191]: There's also a PROT_NONE at line 127 in this file, might have originally been my fault that was introduced.

Comment (jesings) [reply src/trusted/service_runtime/sel_addrspace.c:191]: Also is PROT_RW the same?

File: src/trusted/service_runtime/nacl_syscall_common.c [hunk 7]
```diff
@@ -730,9 +567,8 @@ int32_t NaClSysOpen(struct NaClAppThread  *natp,
     memmove(left_side, right_side, len_of_rest);
   }
 
-  if (retval) {
-    return retval;
-  }
+  if (retval) return retval;
+
```
jesings, line 571: Why this change?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1032,14 +741,8 @@ int32_t NaClSysPread(struct NaClAppThread  *natp, //will make NaCl logs like rea
     count = INT32_MAX;
   }
 
-  NaClVmIoWillStart(nap,
-                    (uint32_t) (uintptr_t) buf,
-                    (uint32_t) (((uintptr_t) buf) + count - 1));
```
jesings, line 1037: How are we able to get rid of these?

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1211,21 +842,10 @@ int32_t NaClSysPwrite(struct NaClAppThread *natp,
   NaClLog(2, "In NaClSysPWrite(%d, %.*s%s, %"NACL_PRIdS")\n",
           d, (int)log_bytes, (char *)sysaddr, ellipsis, count);
 
-  NaClVmIoWillStart(nap,
-                    (uint32_t)(uintptr_t)buf,
-                    (uint32_t)(((uintptr_t)buf) + count - 1));
```
jesings, line 1216: Again, not sure what this was doing and whether it's safe to remove

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -1832,54 +1322,7 @@ int32_t NaClSysMmapIntern(struct NaClApp        *nap,
...
-      map_result = -NACL_ABI_EINVAL;
-      goto cleanup;
-    }
-    if ((NACL_ABI_PROT_WRITE & prot)) {
-      NaClLog(2, "%s\n", "NaClSysMmap: asked for writable and executable code pages?!?");
-      map_result = -NACL_ABI_EINVAL;
-      goto cleanup;
-    }
-    mapping_code = 1;
-  } else if ((prot & NACL_ABI_PROT_EXEC)) {
```
jesings, line 1882: I'd appreciate if you commented that you got rid of this

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -2027,20 +1465,7 @@ int32_t NaClSysMmapIntern(struct NaClApp        *nap,
...
-                 1)) {
-      NaClLog(4, "NaClSysMmap: descriptor not blessed\n");
-      map_result = -NACL_ABI_EINVAL;
-      goto cleanup;
-    }
-    NaClLog(4, "NaClSysMmap: allowed\n");
-  } else if (NaClSysCommonAddrRangeContainsExecutablePages(nap,
+  if (NaClSysCommonAddrRangeContainsExecutablePages(nap,
                                                            usraddr,
                                                            length)) {
```
jesings: Fix indentation

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -2085,214 +1509,57 @@ int32_t NaClSysMmapIntern(struct NaClApp        *nap,
...
-        map_result = (int32_t) usraddr;
-        goto cleanup;
-      }
 
-      NaClLog(3,
-              "NaClSysMmap: did not validate in readonly_text mode;"
-              " attempting to use dyncode interface.\n");
 
-      if (holding_app_lock) {
-        NaClVmHoleClosingMu(nap);
```
jesings, line 2243: I'm not fully sure what this does, please explain so we can make sure this won't cause any vulnerabilities/race conditions

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -2635,7 +1899,7 @@ static int32_t MprotectInternal(struct NaClApp *nap,
             "addr 0x%08"NACL_PRIxPTR", desc 0x%08"NACL_PRIxPTR"\n",
             addr, (uintptr_t) entry->desc);
 
-    if (!entry->desc) {
+    if (!entry->desc && entry->offset == ENTRY_OFFSET_NOFD) { // with no nacl desc or entry offset set to ENTRY_OFFSET_NOFD, this is anonymous
```
jesings: Your comment says this should be an or but it's an and

jesings (approved):
Overall really happy with this and it makes things so much cleaner! Just a couple remaining questions

src/trusted/service_runtime/sel_addrspace.c
rennergade: PROT_RW is defined the same way, yes. the PROT_NONE here is from the original google code it seems (remove histroy commit). Don't think we need to worry about that.

jesings: Ok that's fine

src/trusted/service_runtime/nacl_syscall_common.c
rennergade: just cosmetic, was trying to standardize the look of these functions

src/trusted/desc/nacl_desc_io.c
rennergade: Hmm. I just reverted this back to its pre lind-hacked up state. I'm not sure why this got removed.

rennergade: Okay. Joey added this sometime in between while adding some syscalls it seems. Removing this shouldn't interfere with any functionality since that stuff was all removed as well.

PR #133, opened 2023-02-16
Thread cancel update

rennergade:
This PR fixes problems with cage cancellation and allows exit to exit thread groups

- Adds handler for SIGSEGV from useraddresses in trusted code (ie unmapped addresses reaching I/O)
- Re-factors cancellation for NaClExitThreadGroup to be used for exit as well as cancellation
- Adds system to trap code in trusted/untrusted, and only use pthread_cancel in untrusted space
- Adds points to exit from if trapped in trusted space (along with Rust PR)


src/trusted/service_runtime/linux/nacl_signal.c
```diff
@@ -362,6 +370,9 @@ static void SignalCatch(int sig, siginfo_t *info, void *uc) {
     NaClExit(-sig);
   }
 
+  // Lind: If we segfault on a user address (presumably because it was unmapped between check and use), we can call that an untrusted fault
+  if ((sig == SIGSEGV) && ((uintptr_t)info->si_addr & ~(natp->nap->addr_bits))) is_untrusted = true;
+
```
jesings, line 375: Hang on, did we not differentiate between these before?

src/trusted/service_runtime/linux/thread_suspension.c
```diff
@@ -70,7 +70,15 @@ void NaClAppThreadSetSuspendState(struct NaClAppThread *natp,
...
     if ((state & NACL_APP_THREAD_SUSPENDING) != 0) {
+      struct NaClThread *host_thread;
+      host_thread = &natp->host_thread;
       /* We have been asked to suspend, so wait. */
+      if (lindcheckthread(natp->nap->cage_id, host_thread->tid)) {
+        while(1) {
+            lindsetthreadkill(natp->nap->cage_id, host_thread->tid, false);
+            NaClThreadExit();
+        }
+      }
```
jesings, line 81: Who do we need this loop?

src/trusted/service_runtime/linux/thread_suspension.c
```diff
@@ -218,6 +226,35 @@ static void WaitForUntrustedThreadToSuspend(struct NaClAppThread *natp) {
   }
 }
 
+// Lind: Similar to NaClUntrustedThreadSuspend
+// Here we CaS to trap thread in either trusted or untrusted code,
+void NaClThreadTrapUntrusted(struct NaClAppThread *natp) {
```
jesings: Could you explain the difference between this function and the NaClUntrustedThreadSuspend function?

jesings: In a comment

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -168,6 +168,10 @@ void WINAPI NaClAppThreadLauncher(void *state) {
 
   NaClLog(1, "%s\n", "NaClAppForkThreadLauncher: entered");
 
+  // Lind: we need to enable pthread cancellation to deal with untrusted teardowns 
+  pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, NULL);
+  pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, NULL);
+
```
jesings: Why do we assume that we can cancel here rather than earlier or later?

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -386,7 +391,7 @@ void NaClAppThreadTeardownInner(struct NaClAppThread *natp, bool active_thread)
   #elif NACL_OSX
     NaClClearMachThreadForThreadIndex(thread_idx);
   #endif
-  }
```
jesings, line 389: Fix indentation of above lines

jesings: (not a huge deal but would be nice

src/trusted/service_runtime/nacl_app_thread.c
```diff
@@ -721,6 +715,37 @@ void NaClAppThreadDelete(struct NaClAppThread *natp) {
...
+      child_thread = &natp_child->host_thread;
+
+      lindsetthreadkill(nap->cage_id, child_thread->tid, true);
+      NaClThreadTrapUntrusted(natp_child); // trap the thread in either trusted or untrusted space
+      if (natp_child->suspend_state == (NACL_APP_THREAD_UNTRUSTED | NACL_APP_THREAD_SUSPENDING)) {
+        NaClThreadCancel(child_thread); // we use pthread cancel async since we know were in untrusted code
+        lindsetthreadkill(nap->cage_id, child_thread->tid, false);
+      }
+
+      while (lindcheckthread(natp_child->nap->cage_id, child_thread->tid));
```
jesings, line 747: Why do we need a while

src/trusted/service_runtime/nacl_signal_common.c
```diff
@@ -145,15 +145,10 @@ void NaClSignalHandleUntrusted(struct NaClAppThread *natp,
...
-      NaClUntrustedThreadSuspend(natp, 0);   
-    } else {
-      if (!natp->cage_mainthread->tearing_down) AddToFatalThreadTeardown(natp->cage_mainthread);
-    }
-    // wait here while we get cleaned up
-    while (1);
+
+    if (!natp->nap->tearing_down) AddToFatalThreadTeardown(natp);
+
+    NaClThreadExit();
```
jesings, line 151: Why did we need to wait for cleanup before but not now?

jesings (commented):
I need to wrap my head around the logic more, some more comments would help

src/trusted/service_runtime/nacl_app_thread.c
rennergade: This is the first function running when a thread launches so besides the logging this is basically the first point we can execute code in the thread.

src/trusted/service_runtime/nacl_signal_common.c
rennergade: That behavior was masking a bug I had created before by tearing down the same thread in the handler, and trying to exit twice. This is fixed and straightforward now.

rennergade:
@jesings also added comments here which I believe will be helpful.

jesings:
Ok, looks fine

PR #134, opened 2023-03-23
Thread Suspend CAS remove and  I/O bypass

rennergade:
Two performance improvements that yield 10% improvement on smaller buffers.

1. Circumvent the dispatcher on I/O calls with bypass functions.

2. Remove compare and swap from suspend functions. This only exists because of the Windows VMHole problem which we don't deal with, so we can save a lot of time by just changing the value instead of using CAS.

### DIFF
--- a/src/trusted/service_runtime/linux/thread_suspension.c
+++ b/src/trusted/service_runtime/linux/thread_suspension.c
@@ -240,16 +240,12 @@ void NaClThreadTrapUntrusted(struct NaClAppThread *natp) {
    * taking locks when pthread_kill() takes effect, so we ask the
    * thread to suspend even if it is currently running untrusted code.
    */
-  while (1) {
-    old_state = natp->suspend_state;
-    DCHECK((old_state & NACL_APP_THREAD_SUSPENDING) == 0);
-    suspending_state = old_state | NACL_APP_THREAD_SUSPENDING;
-    if (CompareAndSwap(&natp->suspend_state, old_state, suspending_state)
-        != old_state) {
-      continue;  /* Retry */
-    }
-    break;
-  }
+
+  old_state = natp->suspend_state;
+  DCHECK((old_state & NACL_APP_THREAD_SUSPENDING) == 0);
+  suspending_state = old_state | NACL_APP_THREAD_SUSPENDING;
+  natp->suspend_state = suspending_state);
+
   /*
    * Once the thread has NACL_APP_THREAD_SUSPENDING set, it may not
    * change state itself, so there should be no race condition in this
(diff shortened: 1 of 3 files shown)


src/trusted/service_runtime/linux/thread_suspension.c
```diff
@@ -240,16 +240,12 @@ void NaClThreadTrapUntrusted(struct NaClAppThread *natp) {
...
-        != old_state) {
-      continue;  /* Retry */
-    }
-    break;
-  }
+
+  old_state = natp->suspend_state;
+  DCHECK((old_state & NACL_APP_THREAD_SUSPENDING) == 0);
+  suspending_state = old_state | NACL_APP_THREAD_SUSPENDING;
+  natp->suspend_state = suspending_state;
```
jesings: A comment why there is no race condition here would be helpful!

src/trusted/service_runtime/linux/thread_suspension.c
```diff
@@ -309,17 +299,13 @@ void NaClUntrustedThreadSuspend(struct NaClAppThread *natp,
...
-      continue;  /* Retry */
-    }
-    break;
-  }
+
+  old_state = natp->suspend_state;
+  new_state = old_state & ~(NACL_APP_THREAD_SUSPENDING |
+                            NACL_APP_THREAD_SUSPENDED);
+  DCHECK((old_state & NACL_APP_THREAD_SUSPENDING) != 0);
+  natp->suspend_state = new_state;
```
jesings, line 320: Same as above

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -708,9 +708,9 @@ int32_t NaClSysRead(struct NaClAppThread  *natp,
   /* Lind - we removed the VMIOWillStart and End functions here, which is fine for Linux
    * See note in sel_ldr.h
    */
-  read_result = lind_read(d, (void *)sysaddr, count, nap->cage_id);
+  read_result = quick_read(d, (void *)sysaddr, count, nap->cage_id);
```
jesings, line 714: Please comment what we're doing here

src/trusted/service_runtime/nacl_syscall_common.c
```diff
@@ -798,7 +798,7 @@ int32_t NaClSysWrite(struct NaClAppThread *natp,
   /* Lind - we removed the VMIOWillStart and End functions here, which is fine for Linux
    * See note in sel_ldr.h
    */
-  write_result = lind_write(d, (void *)sysaddr, count, nap->cage_id);
+  write_result = quick_write(d, (void *)sysaddr, count, nap->cage_id);
```
jesings, line 806: Comment!!!!

jesings (approved):
Comments would be helpful!

PR #136, opened 2023-04-28
Remove teardown for SIGPIPE

AlpacaMax:


jesings (approved):
Short and looks good!

PR #139, opened 2023-06-27
Fixed DISPATCH_SYSCALL macro used to invoke flock.

kpan2034:
The arguments for `flock` that Native Client receives were not being passed to `safeposix`. This change fixes that so that the arguments are passed correctly.


jesings (approved):

rennergade:
Minor fix for a dispatcher problem. Looks good.

PR #142, opened 2023-08-28
remove lkm

yizhuoliang:
## Description

Very simple changes, just removing any code involves loadable kernel modules. Most of the LKM related code are depend on the value of `use_lkm`, so I deleted the code being only executed when `use_lkm` is true, and only keep the code being executed when `use_lkm` is false. Besides, just remove the `checkForLKM()` function and it's reference.

### Type of change
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
## How Has This Been Tested?
passed the make test

### DIFF
--- a/src/trusted/service_runtime/nacl_app_thread.c
+++ b/src/trusted/service_runtime/nacl_app_thread.c
@@ -39,7 +39,6 @@
 struct NaClMutex ccmut;
 struct NaClCondVar cccv;
 int cagecount;
-extern bool use_lkm;
 
 struct NaClThread reaper;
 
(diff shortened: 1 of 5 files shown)


rennergade:
@yizhuoliang I know this is a pretty self-explanatory PR but your description should be a bit more than what you added here.

justyoyo123:
I agree with Nick that a more comprehensive explanation as to why these changes were necessary would be helpful. However, the changes themselves look good to me.

Yaxuan-w (approved):
Really a great job! With the addition of more detailed description, the logic of the entire code is well understood. It removes lkm related code and avoids future confusion. This change looks good to me, so approved.

kuzeyardabulut (approved):
Fantastic work! The additional detailed description significantly enhances the clarity of the entire codebase. By removing the lkm-related code, you've effectively eliminated any potential confusion going forward.

PR #158, opened 2024-03-09
Fix Program Counter on Fork

rennergade:
## Description

This fixes a segfault in situations where a child cage is forked and the parent immediately exists. Here the child program counter isnt properly adjusted and still points to the parents address space. Because the parent frees its address space, the program faults. It's pretty crazy that this one took this long to catch, but because of how NaCl is setup, as long as it enters the code correctly the addresses are then properly located after the point of entry which made this tricky to find.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

All test suites and applications

## Checklist:
- [x] My code follows the style guidelines of this project
- [x ] My changes generate no new warnings


yizhuoliang (approved):
I see how crazy it is. This change makes sense.

RusherRG (approved):
Awesome. We finally have this issue fixed.

yzhang71 (approved):
LGTM! Thanks for the fix.

